PR #19, opened 2016-12-17
docs: mention access token scope in 'deployment'

Charlotteis:
I mentioned the access scope here again because I skipped right to the end to learn how to deploy, and folks might not read 'development'.


andrew:
👍 want to add a note to the ENV description in app.json as well?

PR #30, opened 2016-12-18
moved heroku deployment on readme.md

GiacomoLaw:
Moved it to the top to give it a little more attention. It's easier than hosting it locally, and moving it up allows more people to use it easily. 😄


andrew:
Good thinking!

PR #31, opened 2016-12-18
Removed annoying gif from all clear view

GiacomoLaw:
I removed the gif from the clear view, and just put text in instead. Looks like this: https://gyazo.com/9b18012d3b491ed1bc804ddf6da3b5e4

I left the gif in the assets though. I just prefer it plain like this, and others may too!

I'll look into adding an option for the gif or for text, but I'm a little bit of a noob so I'm not really sure how to do that. 😄


GiacomoLaw:
You can see it here: https://github-inbox-giacomolaw.herokuapp.com/

andrew:
Let's remove the gif from the assets folder too

GiacomoLaw:
Done 😉 

Thanks for the awesome tool!

PR #35, opened 2016-12-18
Add support for Docker Compose development environment

hectcastro:
Include the Docker configuration files necessary to support a PostgreSQL backed instance of the application running on Ruby 2.3.

Fixes https://github.com/andrew/octobox/issues/12

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,4 +3,5 @@ FROM ruby:2.3-onbuild
 RUN \
     apt-get update \
     && apt-get install -y --no-install-recommends nodejs \
+    && apt-get clean \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM ruby:2.3-onbuild
+
+RUN \
+    apt-get update \
+    && apt-get install -y --no-install-recommends nodejs \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ rails s
 ```
 #### Docker Compose
 
-If you're familiar with Docker and Docker Compose, the included `docker-compose.yml` configuration allows you to spin up the application locally.
+If you're familiar with [Docker](https://docs.docker.com/engine/) and [Docker Compose](https://docs.docker.com/compose/), the included `docker-compose.yml` configuration allows you to spin up the application locally.
 
 First, launch an instance of PostgreSQL and wait for it to fully initialize:
 
--- a/README.md
+++ b/README.md
@@ -103,6 +103,8 @@ Once the PostgreSQL initialization process is complete, launch the application u
 GITHUB_TOKEN=yourpersonalaccesstoken docker-compose up app
 ```
 
+**Note**: You can add `GITHUB_TOKEN` to `.env` instead of supplying it directly on the command-line.
+
 ### Note on Patches/Pull Requests
 
  * Fork the project.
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    environment:
+      - POSTGRES_PASSWORD=development
+
+  app:
+    image: github-inbox
+    build:
+      context: ./
+      dockerfile: Dockerfile
+    ports:
+      - 3000:3000
```
WillAbides: It would be nice to see something like `- ${GITHUB_INBOX_PORT:-3000}:3000` here, but that requires using version 2.1 of the docker-compose.yml format, which in turn requires docker-compose >= 1.9.  I'm not sure if enough people are on 1.9 yet.

andrew:
I've just [renamed](https://github.com/andrew/octobox/pull/38) the project to Octobox, can you update the database and ENV config names?

Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ruby:2.3-onbuild
+
+RUN \
+    apt-get update \
+    && apt-get install -y --no-install-recommends nodejs \
```
calavera: you might want to use `apt-get clean` here to remove archives from the apt cache:

```
clean
    clean clears out the local repository of retrieved package files. It removes everything but the lock file from /var/cache/apt/archives/ and /var/cache/apt/archives/partial/
````

File: docker-compose.yml (hunk 1 above)
Comment (hectcastro) [inline docker-compose.yml]: I'm leaning toward leaving it as is to that things will work for a wider audience, but maybe @andrew can weigh in.

(I did not know about Docker Compose 1.9, or the update to this behavior in the Docker Compose YAML format, so thanks for the heads up about it.)

File: Dockerfile (hunk 2 above)
Comment (hectcastro) [inline Dockerfile]: Addressed in 47a499af66ac09238544a050d9cc2e6c25e65d3a.

File: docker-compose.yml (hunk 1 above)
Comment (WillAbides) [inline docker-compose.yml]: I think you're right.  If 1.9 was a few months older, it would be worth doing.  Maybe in April...

**Comment by hectcastro** [comment]
I think I've addressed all of the comments.

Regarding the `docker-compose` commands for `database` and `app` in the `README`, they're split to avoid the Rails application from attempting to connect to the PostgreSQL database before it is completely initialized.

**Comment by andrew** [comment]
Thanks @hectcastro 🎉

File: README.md [hunk 3]
```diff
@@ -87,7 +87,23 @@ Finally you can boot the rails app:
 ```bash
 rails s
 ```
+#### Docker Compose
 
+If you're familiar with [Docker](https://docs.docker.com/engine/) and [Docker Compose](https://docs.docker.com/compose/), the included `docker-compose.yml` configuration allows you to spin up the application locally.
```
bfirsh: Probably best to link to: https://docs.docker.com/installation/

Docker is actually of more value if you're not familiar with it. My normal instructions here are:

Step 1: [install Docker for Mac/Windows.](https://docs.docker.com/installation/)
Step 2: run `docker-compose up`

Boom.

README.md
```diff
@@ -87,7 +87,23 @@ Finally you can boot the rails app:
...
+
+First, launch an instance of PostgreSQL and wait for it to fully initialize:
+
+```bash
+docker-compose up database 
+```
+
+Once the PostgreSQL initialization process is complete, launch the application using another terminal session:
```
bfirsh: You can use netcat to avoid this dance: http://stackoverflow.com/questions/31746182/docker-compose-wait-for-container-x-before-starting-y

README.md
andrew: Thanks for the tips @bfirsh, I don't suppose fancy sending the changes as a pr?

bfirsh: 👍 will have a look later

PR #40, opened 2016-12-18
Put a check in the archive checkbox if the notification is archived

WillAbides:
Changes the archive/unarchive checkbox to be checked when the notification is archived.

2016-12-18 at 4 20 pm


andrew:
👍

Issue #13, opened 2016-12-17
Add a contributing.md


andrew:
Adapted from the excellent one in 24pullrequests: https://github.com/24pullrequests/24pullrequests/blob/master/CONTRIBUTING.md

PR #41, opened 2016-12-18
adding CONTRIBUTING.md

nateprewitt:
This should address #13 with adding a CONTRIBUTING.md modeled after the one from 24PullRequests.

One thing that's outstanding here is whether this project intends to support Vagrant. The README seems to suggest that it will, but there isn't a supplied Vagrantfile. If there's intent to support this in the future, I can leave that Vagrant section in, otherwise I'll remove it.


andrew:
Thanks for picking this up @nateprewitt, I've just merged some [docker support](https://github.com/andrew/octobox/pull/35) so we probably won't need the vagrant stuff so you can take it out.

nateprewitt:
Great, will do.

andrew:
Is this ready for merging now?

nateprewitt:
Yeah sorry, should be good to go.

andrew:
No probs, GitHub doesn't notify anyone when you force push, thanks!

PR #47, opened 2016-12-19
Add 'Archive All' button

pezholio:
While fiddling about, I thought this may be useful to nuke all notifications and start from scratch. Influenced by the Gmail interface, I've put the button under a 'More' submenu to prevent it from being clicked accidentally.

I've also added `factory_girl` as a test dependency to make creating database-level objects a bit easier.

### DIFF
--- a/test/factories/user.rb
+++ b/test/factories/user.rb
@@ -1,7 +1,7 @@
 FactoryGirl.define do
   factory :user do
-    github_id 1234
-    access_token '54c543ffd45343fg5434fgg556ff553232gg57890v'
+    github_id { rand(0..5000) }
+    access_token { SecureRandom.hex(15) }
     github_login 'andrew'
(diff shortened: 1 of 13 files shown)


test/factories/user.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :user do
+    github_id 1234
+    access_token '54c543ffd45343fg5434fgg556ff553232gg57890v'
```
tarebyte: Since there are unique indexes on `github_id` and `access_token` columns I'm pretty sure the linting for FactoryGirl will fail for the user.

pezholio: Ah yeah, I'll bin that off. I'm using random generators for those now

andrew:
Noice 👌

PR #57, opened 2016-12-19
Only fetch updated notifications since the last sync

pixeltrix:
If you have a lot of notifications then downloading them all every time is going to be painful and probably blow the rate limit on the API.

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -13,7 +13,57 @@ class Notification < ApplicationRecord
   scope :status,   ->(status)       { where(unread: status) }
 
   paginates_per 20
-  
+
+  class << self
+    def download(user)
+      timestamp = Time.current
+
+      fetch_notifications(user) do |notification|
+        begin
+          n = find_or_create_by(github_id: notification.id)
+
+          if n.archived && n.updated_at < notification.updated_at
+            n.archived = false
+          end
+
+          attrs = {
+            user_id: user.id,
+            repository_id: notification.repository.id,
+            repository_full_name: notification.repository.full_name,
+            subject_title: notification.subject.title,
+            subject_url: notification.subject.url,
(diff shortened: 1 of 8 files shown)


app/models/notification.rb
```diff
@@ -13,7 +13,56 @@ class Notification < ApplicationRecord
...
+            subject_type: notification.subject.type,
+            reason: notification.reason,
+            unread: notification.unread,
+            updated_at: notification.updated_at,
+            last_read_at: notification.last_read_at,
+            url: notification.url
+          }
+
+          n.update(attrs)
+        rescue ActiveRecord::RecordNotUnique => e
```
tarebyte: Nice 👍

pixeltrix:
Done 👍

pixeltrix:
Hmm, don't merge it yet - just blew up locally

andrew:
👍 We should probably also add a limit on the initial sync of perhaps just notifications from the past week/month to avoid download potentially huge amounts of old notifications.

pixeltrix:
Okay, enhanced the test to actually create notifications and we only download a months worth on first sync

test/models/notification_test.rb
```diff
@@ -1,7 +1,33 @@
 require 'test_helper'
 
 class NotificationTest < ActiveSupport::TestCase
-  # test "the truth" do
-  #   assert true
-  # end
+  setup do
+    notifications_url = %r{https://api.github.com/notifications}
+
+    body     = File.read(Rails.root.join('test', 'fixtures', 'files', 'notifications.json'))
```
pixeltrix: Possibly - I should know about that shouldn't I? 😊

pixeltrix: Done

andrew:
🚢

PR #58, opened 2016-12-19
Automatically sync notifications after signup

andrew:
Fixes https://github.com/andrew/octobox/issues/49


andrew:
Yeah, just trying to remember how to fire the after_commit callbacks in tests 🙈

andrew:
@tarebyte any tips on how to test this?

tarebyte:
I would just create an object and make an assertion that the API endpoint was called.

andrew:
I've pushed a failing test, it's midnight here and my brain is broken!

test/models/user_test.rb
```diff
@@ -63,4 +63,18 @@ class UserTest < ActiveSupport::TestCase
       assert_equal n.archived, true
     end
   end
+
+  test "triggers sync_notifications on save" do
+    notifications_url = %r{https://api.github.com/notifications}
+
+    body     = '[]'
+    headers  = { 'Content-Type' => 'application/json' }
+    response = { status: 200, body: body, headers: headers }
```
tarebyte: This is starting to become a pattern, we may want to look into a helper for this.

tarebyte:
@andrew so the `after_commit` ActiveRecord callback was making the call recursive so it was an infinite loop 😆 

I've updated it to use `after_create` which works and has the same functionality ¯\\_(ツ)_/¯

andrew:
Ahhh that would explain it, thanks for the sanity check!

PR #62, opened 2016-12-20
Make sure that a Repository Invitation notification has a URL

tarebyte:
Right now if you receive a Repository Invitation notification Octobox will break because the `subject_url` doesn't exist.

This PR updates fills in the `subject_url` to link to the repository invitation so that things don't break.


tarebyte:
Also apparently we don't send Organization Invitation notifications so I think we should be good!

andrew:
👍

PR #63, opened 2016-12-20
open current notification with Enter key

### LINKED ISSUE #61: Open notifications via keyboard shortcut
It would be nice if pressing `o` or `enter` would open the current notification link in a new tab.

karpiu:
Fixes #61


andrew:
Brilliant, thanks!

Issue #59, opened 2016-12-19
UI looks funky after archiving everything


jdm:
<img width="1386" alt="screen shot 2016-12-19 at 6 18 49 pm" src="https://cloud.githubusercontent.com/assets/27658/21332610/a1c7dd84-c617-11e6-8c88-df565b39ba16.png">

karpiu:
Don't know about "funky" but I noticed a bug in your screenshot that is probably my fault: the little blue vertical line for navigating notifications. It should not be visible if the list is empty. I will fix it.

tarebyte:
Closed by #66

karpiu:
@tarebyte Not sure if you should close this one. @jdm have not explained yet what "funky" means

jdm:
It was the blue line, but also the big white space on the top left above the Archive link.

andrew:
Blue line is fixed, big white space on the left needs to be squashed as there are empty tables there padding it out

karpiu:
@andrew Ok. I will fix this.

PR #66, opened 2016-12-20
moving messeges about empty list outside of the notifications table

karpiu:
Fixes bug mentioned in #59


andrew:
Thanks for all your help @karpiu!

PR #72, opened 2016-12-20
Add support for GitHub Enterprise

### LINKED ISSUE #14: GitHub Enterprise config
In theory this should "just work" with an enterprise install if we make the domain configurable in the Notification class

tarebyte:
Resolves #14 by adding support for GitHub Enterprise by allowing the user to set the `DOMAIN` environment variable.

Not sure about the best way to test this so I'd ❤️ some suggestions 😄

/cc @andrew

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,18 @@ You can host your own instance of Octobox using Heroku. Heroku will ask you to p
 
 [![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy)
 
+## Running Octobox for [GitHub Enterprise](https://enterprise.github.com/home)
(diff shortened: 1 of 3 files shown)


andrew:
Will it always be github.foobar.com or could it be gh.foobar.com? I've never setup an enterprise install so not sure of the rules.

Also perhaps we should call the ENV var `GITHUB_DOMAIN`?

tarebyte:
> Will it always be github.foobar.com or could it be gh.foobar.com? I've never setup an enterprise install so not sure of the rules.

Yeah know what, neither have I 😆

I'll update it so they set the entire domain so we're not guessing.

andrew:
👍

PR #74, opened 2016-12-20
store reposiroty owner name in database

### LINKED ISSUE #27: Store owner name on notification
To allow easy filtering and grouping by user/org that owns the repo we should add an `owner_name` field to the notifications table and store it when downloading notifications.

karpiu:
Fixes #27

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -30,6 +30,7 @@ class Notification < ApplicationRecord
             attr[:user_id]              = user.id
             attr[:repository_id]        = notification.repository.id
             attr[:repository_full_name] = notification.repository.full_name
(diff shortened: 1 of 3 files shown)


app/models/notification.rb
```diff
@@ -30,6 +30,7 @@ def download(user)
             attr[:user_id]              = user.id
             attr[:repository_id]        = notification.repository.id
             attr[:repository_full_name] = notification.repository.full_name
+            attr[:repository_owner_name]= notification.repository.owner.login
```
tarebyte: Mind helping my inner Rubocop 🤖 and fix the spacing here? Thanks!

karpiu: @tarebyte Sure :smile_cat:

PR #77, opened 2016-12-20
target changed to _blank for notification links

### LINKED ISSUE #71: Opening notifications overrides the previous opened notification
If I press enter on three notifications in a row, only a single tab is opened and it shows the content of the last notification that I opened. I would expect this to open three separate tabs, instead.

karpiu:
Fixes #71


andrew:
👌

PR #81, opened 2016-12-20
15 mobile friendly fixes

briannelson:
Fairly opinionated changes here, but helps a lot on mobile devices and general legibility

octobox

octobox

octobox

For #15

PR #84, opened 2016-12-20
Make it easier to obtain the GitHub domain

tarebyte:
Will help with #81 

Let's make it more straight forward as to how to get the GitHub domain that the app is pointing to.

/cc @andrew


andrew:
Ah good thinking!

PR #90, opened 2016-12-21
Refactor Complex Methods to Lower Cyclomatic Complexity

michaelrauh:
This is a small refactor that removes some of the branching from the methods that are flagged by Rubocop for having high cyclomatic complexity. The subject_type method was not flagged, but following the pattern from the refactor for reason_name seemed easier to follow.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -13,13 +13,17 @@ class NotificationsController < ApplicationController
     @unread_repositories = scope.distinct.group(:repository_full_name).count
     @starred             = scope.starred.count
 
-    scopes = {params[:repo] => scope.repo(params[:repo]), params[:reason] => scope.reason(params[:reason]),
-       params[:type] => scope.type(params[:type]), params[:status] => scope.status(params[:status]),
-       params[:starred] => scope.starred}
+    scopes = {
+      params[:repo]    => scope.repo(params[:repo]),
+      params[:reason]  => scope.reason(params[:reason]),
+      params[:type]    => scope.type(params[:type]),
+      params[:status]  => scope.status(params[:status]),
+      params[:starred] => scope.starred
+    }
 
-    scopes.each do |k, v|
-      if k.present?
-        scope = v
+    scopes.each do |param_value, sub_scope|
+      if param_value.present?
+        scope = sub_scope
       end
     end
 
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -4,8 +4,8 @@ class NotificationsController < ApplicationController
   before_action :render_home_page_unless_authenticated, only: [:index]
 
   def index
-    scope = current_user.notifications
-    scope = params[:archive].present? ? scope.archived : scope.inbox
+    notifications = current_user.notifications
+    scope = params[:archive].present? ? notifications.archived : notifications.inbox
 
     @types               = scope.distinct.group(:subject_type).count
     @statuses            = scope.distinct.group(:unread).count
@@ -13,11 +13,15 @@ class NotificationsController < ApplicationController
     @unread_repositories = scope.distinct.group(:repository_full_name).count
     @starred             = scope.starred.count
 
-    scope = scope.repo(params[:repo])     if params[:repo].present?
-    scope = scope.reason(params[:reason]) if params[:reason].present?
-    scope = scope.type(params[:type])     if params[:type].present?
-    scope = scope.status(params[:status]) if params[:status].present?
-    scope = scope.starred                 if params[:starred].present?
+    scopes = {params[:repo] => scope.repo(params[:repo]), params[:reason] => scope.reason(params[:reason]),
+       params[:type] => scope.type(params[:type]), params[:status] => scope.status(params[:status]),
+       params[:starred] => scope.starred}
+
+    scopes.each do |k, v|
+      if k.present?
+        scope = v
+      end
+    end
 
     @notifications = scope.newest.page(params[:page])
   end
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -14,34 +14,16 @@ def flash_messages
...
-    when 'RepositoryInvitation'
-      'mail-read'
-    when 'Issue'
-      'issue-opened'
-    when 'PullRequest'
-      'git-pull-request'
-    when 'Commit'
-      'git-commit'
-    end
+    subject_types = {'RepositoryInvitation' => 'mail-read', 'Issue' => 'issue-opened',
```
tarebyte: So instead of defining the hash here, what do you think about nesting it at the top of the module and [freezing](http://ruby-doc.org/core-2.3.3/Object.html#method-i-freeze) it?

```ruby
module ApplicationHelper
  SUBJECT_TYPES = {
    'RepositoryInvitation' => 'mail-read',
    'Issue'                => 'issue-opened',
    'PullRequest'          => 'git-pull-request',
    'Commit'               => 'git-commit'
  }.freeze

  def notification_icon(subject_type)
    SUBJECT_TYPES[subject_type]
  end
end
```

app/helpers/application_helper.rb
```diff
@@ -14,34 +14,16 @@ def flash_messages
...
-    when 'mention'
-      'warning'
-    when 'assign'
-      'danger'
-    when 'subscribed'
-      'default'
-    else
-      'default'
-    end
+    reasons = {'comment' => 'primary', 'author' => 'success', 'state_change' => 'info', 'mention' => 'warning',
```
tarebyte: Same comment here as above, could be something like.

```ruby
module ApplicationHelper
  REASON_LABELS = {
    'comment'      => 'primary',
    'author'       => 'success',
    'state_change' => 'info',
    'mention'      => 'warning',
    'assign'       => 'danger'
  }.freeze

  def reason_label(reason)
    REASON_LABELS.fetch(reason, 'default')
  end
end
```

And for the `default` you could just use just use [fetch](https://ruby-doc.org/core-2.3.3/Hash.html#method-i-fetch)

app/controllers/notifications_controller.rb
```diff
@@ -4,20 +4,24 @@ class NotificationsController < ApplicationController
   before_action :render_home_page_unless_authenticated, only: [:index]
 
   def index
-    scope = current_user.notifications
-    scope = params[:archive].present? ? scope.archived : scope.inbox
+    notifications = current_user.notifications
```
tarebyte: Technically this is still a scope because ActiveRecord doesn't query the data until we ask it to iterate over or return the list of objects. So I don't quite agree with this name change.

app/controllers/notifications_controller.rb
```diff
@@ -4,20 +4,24 @@ class NotificationsController < ApplicationController
...
     @reasons             = scope.distinct.group(:reason).count
     @unread_repositories = scope.distinct.group(:repository_full_name).count
     @starred             = scope.starred.count
 
-    scope = scope.repo(params[:repo])     if params[:repo].present?
-    scope = scope.reason(params[:reason]) if params[:reason].present?
-    scope = scope.type(params[:type])     if params[:type].present?
-    scope = scope.status(params[:status]) if params[:status].present?
-    scope = scope.starred                 if params[:starred].present?
+    scopes = {params[:repo] => scope.repo(params[:repo]), params[:reason] => scope.reason(params[:reason]),
```
tarebyte: I like where this is going, but it's really terse.

Can you break this up into a multiline hash?

app/controllers/notifications_controller.rb
```diff
@@ -4,20 +4,24 @@ class NotificationsController < ApplicationController
...
+       params[:type] => scope.type(params[:type]), params[:status] => scope.status(params[:status]),
+       params[:starred] => scope.starred}
+
+    scopes.each do |k, v|
```
tarebyte: I understand what it does, but would you mind making the variables names a little more explicit?

app/controllers/notifications_controller.rb
michaelrauh: That makes sense. I only changed the name because scope was being assigned twice in a row here without being read. Is there a name for the variable notifications that would make more sense to you?

app/controllers/notifications_controller.rb
michaelrauh: Yes, I will break this up and freeze it

app/controllers/notifications_controller.rb
michaelrauh: Oops! I'll get on that

michaelrauh:
@tarebyte Do these changes make more sense to you? Thank you for the feedback!

tarebyte:
> Do these changes make more sense to you?

They do, I'll think over the notification variable name.

Thanks for the contribution 🎉

app/controllers/notifications_controller.rb
```diff
@@ -4,20 +4,28 @@ class NotificationsController < ApplicationController
...
     @reasons             = scope.distinct.group(:reason).count
     @unread_repositories = scope.distinct.group(:repository_full_name).count
     @starred             = scope.starred.count
 
-    scope = scope.repo(params[:repo])     if params[:repo].present?
-    scope = scope.reason(params[:reason]) if params[:reason].present?
-    scope = scope.type(params[:type])     if params[:type].present?
-    scope = scope.status(params[:status]) if params[:status].present?
-    scope = scope.starred                 if params[:starred].present?
+    scopes = {
```
tarebyte: Looking at this change again, this doesn't technically work. Here's why:

The way we were doing it before was that we were adding another scope onto the query for each parameter.

This just keeps replacing the scope instead of building on top of it.

michaelrauh: Good catch! I have reverted the changes to this file.

tarebyte:
I'll let @andrew give another 👀 over this before merging.

andrew:
Nice!

PR #101, opened 2016-12-21
Make the filter dots better

briannelson:
I assigned the wrong colors to the original dots for Author, Mention, Comment, etc.

This makes the dots a bit bigger and the colors correspond directly to the label colors used in the message list.

octobox_and_octobox_and_briannelson_github-inbox

Before / After


andrew:
👌

briannelson:
Squeezed one in right before merging, I guess, though it shouldn't break anything :)

andrew:
Sneaky!

PR #108, opened 2016-12-21
Add link to inbox in the sidebar

knolleary:
When in the Archive view, I could not find anyway to get back to the normal view.

*edit*: just found the `refresh` button does it... but that wasn't obvious to me.

This PR adds an 'Inbox' link in the sidebar to help the user navigate between the views.


GiacomoLaw:
Hit Octobox at the top left.

andrew:
👍

PR #110, opened 2016-12-21
navigation cursor remembers its position

### LINKED ISSUE #70: Cursor position resets after archiving
If I leave several notifications to look at later and want to archive a bunch of ones below them, this means I have to scroll down past them every time. This feels inefficient to me!

karpiu:
fixes #70

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -15,14 +15,14 @@ document.addEventListener("turbolinks:load", function() {
     $(this).toggleClass("star-active star-inactive")
     $.get('/notifications/'+$(this).data('id')+'/star')
   });
-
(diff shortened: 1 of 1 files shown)


tarebyte:
@karpiu I'm seeing some weird jumping of the cursor that I don't think if quite right. What do you think?

jumpy

karpiu:
@tarebyte I will check it tomorrow.

karpiu:
@tarebyte the problem is that turbolinks is caching previously visited pages. I added a function that clears the cursor before page is cached. You still might experience a little flicker on a cursor but now it should be only in place. Is that acceptable?

andrew:
🚢

PR #113, opened 2016-12-21
Archive all Notifications within the context of the page

### LINKED ISSUE #75: Archive all should be context specific
If I am viewing all issues from a specific repository, I want to clear them all in one click.
Archive all hits everything in all my notifications.

Is there currently a way to do what I want?

tarebyte:
Closes #75

This add the URL parameters to the Archive all post action so that we only archive in the context of what we are seeing.

I'm not super happy with this solution as I think we can DRY this up a bit, so I'd ❤️  feedback.

I also removed the Archive all button from the Archive page.

/cc @andrew

### DIFF
--- a/app/views/notifications/index.html.erb
+++ b/app/views/notifications/index.html.erb
@@ -56,8 +56,20 @@
           <%= octicon 'sync', :height => 16 %>
         <% end %>
 
-        <% if @notifications.any? %>
-          <%= link_to 'Archive all', archive_notifications_path, class: 'btn btn-default', method: :post %>
+        <% if @notifications.any? && !params[:archive].present? %>
+          <%= link_to(
+                'Archive all',
+                archive_notifications_path(
+                  repo:    params[:repo],
+                  reason:  params[:reason],
+                  type:    params[:type],
+                  status:  params[:status],
+                  starred: params[:starred]
+                ),
+                class: 'btn btn-default',
+                method: :post
+              )
+          %>
         <% end %>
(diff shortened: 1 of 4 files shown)


andrew:
A good start to stop people accidentally archiving everything, feels like we can replace it with a "select all" checkbox and a single archive button the archives all the selected notifications

tarebyte:
> A good start to stop people accidentally archiving everything, feels like we can replace it with a "select all" checkbox and a single archive button the archives all the selected notifications

Agreed, I don't think this is the end all solution. But in the mean time it will save users some confusion.

andrew:
Merge away!

app/views/notifications/index.html.erb
```diff
@@ -56,8 +56,20 @@
           <%= octicon 'sync', :height => 16 %>
         <% end %>
 
-        <% if @notifications.any? %>
-          <%= link_to 'Archive all', archive_notifications_path, class: 'btn btn-default', method: :post %>
+        <% if @notifications.any? && !params[:archive].present? %>
+          <%= link_to(
+                'Archive all',
```
briannelson: @tarebyte change wording to be more clear?

"Archive this list" ?

tarebyte:
@briannelson I've updated the view so that if are actually archiving all of your notifications it will say so. Otherwise it will say `Archive this list`. Thanks for the suggestion!

andrew:
👍

Issue #53, opened 2016-12-19
Show current status of issue/PR


davejlong:
In the notification list, I found myself wanting a way to quickly see if an issue was closed or a PR was merged, instead of having to click through each notification to check.

davejlong:
Closing per #4

PR #115, opened 2016-12-21
Add filter breadcrumbs in top toolbar when filters are active

knolleary:
I stumbled over two issues with the list of filters:

1. it wasn't, at a glance, obvious which filters were active
2. you cannot disable a filter without resetting the lot

This PR adds breadcrumbs in the toolbar when filters are active. Clicking on any of the breadcrumbs removes it from the active filter list. I see this as a precursor to having a Search box (#53) in the gmail style that exposes the filters as search terms... but one thing at a time. 

Looks like this:

cursor_and_octobox


andrew:
👌

tarebyte:
Ooo I like this @knolleary!! I we might be able to clean this view up a bit with a reusable component.

tarebyte:
Looking at this again, they are rather unique. @knolleary could I convince you to make this whole section it's own partial?

knolleary:
@tarebyte yeah, not very DRY in there. Wanted to share the concept with a working implementation before spending too much time on it.

Happy to refine it - when you say 'whole section' I assume you mean the Filter box rather than the toolbar as a whole? Will have a play.

tarebyte:
> I assume you mean the Filter box rather than the toolbar as a whole?

Yup that's correct, apologies for not being more explicit.

knolleary:
Have moved the filter breadcrumbs (for want of a better name) to their own partial. Given how much repetition there is in there, it feels like the filter button ought to be its own partial and we iterate over a list of active filters... but that goes beyond my current understanding of rails/erb to achieve.

tarebyte:
@knolleary I'd be happy to help out if you enable maintainers to push to your branch 😄

knolleary:
@tarebyte thought I had... add_filter_breadcrumbs_in_top_toolbar_when_filters_are_active_by_knolleary_ _pull_request__115_ _andrew_octobox

tarebyte:
Yup! I just missed it on the PR page 😄

tarebyte:
@knolleary I've made some changes, let me know what you think!

knolleary:
Looks good.

I see you've removed `starred` as a filter ... is that a consequence of #116?

tarebyte:
Yup!

knolleary:
Hmm, not sure about that. To me, 'starred' is just another filter on the view of notifications. If you want to see all starred, you just apply the starred filter. To see starred in a particular repo that are issues... You pick that combination of filters.

knolleary:
Having resolved #119, I think we just need to restore the starred filter here and this will be good to merge.

andrew:
This gem might make managing some of these combinable filters nicer: https://github.com/marcelocf/searrrch

> Allow you to receive a query string using the syntax similar to gmail's filters and parse it without going crazy.

> The following search string is supported:

    user_id: 123 user_name: omg user_id: 124 free text here
> The user can also put the value between quotes, in which case nested escaped quotes are supported.

knolleary:
Have added the starred filter back in.

@andrew - that does look useful. Can we look at it in a separate PR as it's a much broader rework of the filters in general?

Tried the `Resolve conflicts` button above, but I'm getting and empty edit box with spinning octocat  and 404 errors in the js console ... @tarebyte

andrew:
Yeah another PR sounds good 👍

knolleary:
Conflicts resolved and I think I'm done here.

Issue #114, opened 2016-12-21
Sidebar filter order


briannelson:
Presumably, Inbox, Archive, and Starred are the most important filters. We should move these to the top.

![octobox](https://cloud.githubusercontent.com/assets/29120/21406334/0f9c7284-c799-11e6-8c95-2c2187512650.png)

tarebyte:
@briannelson I was just thinking that as well, feel free to open a PR!

briannelson:
I'll try to work on a better sidebar display tonight. Too many lines for my taste :) No offense @andrew

andrew:
I'm immune to offense when it comes to design work I did on Sunday night!

knolleary:
Just raised PR #115 related to filters.
Whilst putting that together, I noticed that the filter links in the sidebar are slightly inconsistent in their behaviour - for example, clicking a `type` filter also clears any current `reason` filter. I didn't touch it under #115 as I didn't want to overload it.

@briannelson - if you're delving in there, could you look at that as well?

briannelson:
I'm actually seeing something that I think is related, but good point @knolleary. Because of the way filters get applied, it's possible to have more than one active at a time, currently. This is more of a UX question @andrew / @tarebyte. 

<img width="411" alt="octobox" src="https://cloud.githubusercontent.com/assets/29120/21407217/66b09402-c79d-11e6-9fa2-e58c5ca9023f.png">

If I'm reading it correctly, there are 2 buckets (Inbox, Archive), and within those you can filter by all the rest. (Starred seems like it should be its own bucket, fwiw, independent of Inbox and Archive). Anyways, not sure a good direction here, just wanted to get it down for now.

tarebyte:
@briannelson yes that is correct, `Starred` is a kind of filter of I don't think it belongs in the same section as the Archive and the Inbox.

briannelson:
So, to be clear, Archive Starred and Inbox Starred are two different lists, potentially?

tarebyte:
> So, to be clear, Archive Starred and Inbox Starred are two different lists, potentially?

They are two different lists correct. The inbox removes the archived parameter from the URL.

briannelson:
That seems like an anti-pattern to me. 
i.e. if I star something, it's important to me whether it's archived it's not. 

My suggestion is to make starred independent of Inbox and Archive so you click it and see ALL your starred messages. The checkbox would be checked for those that are archived and unchecked for those that aren't.

[...]

PR #116, opened 2016-12-21
Let Stars list show both Archived and Unarchived notifications

tarebyte:
From #114 this lets the Starred column be it's own list.

screen shot 2016-12-21 at 17 42 04

/cc @briannelson


app/controllers/notifications_controller.rb
```diff
@@ -4,20 +4,24 @@ class NotificationsController < ApplicationController
   before_action :render_home_page_unless_authenticated, only: [:index]
 
   def index
-    scope = current_user.notifications
-    scope = params[:archive].present? ? scope.archived : scope.inbox
+    scope    = current_user.notifications
+    @starred = scope.starred.count
+
+    scope = if params[:starred].present?
```
michaelrauh: I think this may be slightly more readable as an if, else if instead of an if with an embedded ternary.

```
if params[:starred].present?
	scope = scope.starred
elsif params[:archive].present?
	scope =  scope.archived
else
	scope = scope.inbox
end
```

The obvious downside is the repetition of `scope=`

briannelson:
Seems like this one is a bit of a point of contention. Let me take a look at some other inbox examples. Just seems that "Important" or "Starred" should look across "containers" like Inbox/Archive.

This may need a broader discussion around the intended purpose of the Inbox/Archive functionality. I'll defer to you guys on this since you're more inundated with notifications than I am.

app/controllers/notifications_controller.rb
tarebyte: I could do something like this:

```ruby
scope = if params[:starred].present?
          scope.starred
        elsif params[:archive].present?
          scope.archived
        else
          scope.inbox
        end

Comment (michaelrauh) [inline app/controllers/notifications_controller.rb]: Option I didn't think of FTW

**Comment by andrew** [comment]
Just having a play around with the current behaviour in master, it does feel odd to have two different views of stars, I think I'd prefer to see all the stars, I think that's similar to what @briannelson is saying.

n.b Little conflict to fix

**Comment by andrew** [comment]
Having read through the code I am 💯  on this pr!

**Comment by tarebyte** [comment]
> Having read through the code I am 💯 on this pr!

@andrew just so I understand are you changing your mind from your previous https://github.com/andrew/octobox/issues/119#issuecomment-268748756

**Comment by andrew** [comment]
Ah mixed issues there, I agree with this part of what @knolleary said, which I think matches with this change:

> From a UX perspective, the 'Starred' link in the sidebar could set the filter to "is:starred" and clear any existing "in:inbox" / "in:archive" filter

I think we should be able to filter by both starred and archived but the way the UI works at the moment is wonky, 

I think that clicking on the "Starred" link in the sidebar should show all starred items by default, and in future we'll want to be able to add an "is:starred" filter to an existing view or search

**Comment by briannelson** [comment]
I am with you guys 100% on that. Think of the 3 top sidebar choices (Inbox, Archived, Starred) as "quick links" to applying that (and only that) filter.

If that's the case, we'll probably want to remove the separator between Archived and Starred and remove the `class: 'filter'` on starred, so it feels more like Inbox and Archived.

**Comment by tarebyte** [comment]
Ok so here's what I did:

I've removed the starred filter for now in lieu of the hotlink, we can follow up and get the hotlink plus the filter to work in another PR.

screen shot 2016-12-22 at 14 38 43

If this is not what we wanted I apologize I'm starting lose track of what the final decision it.

:shipit: @andrew ?

File: app/views/notifications/_sidebar.html.erb [hunk 2]
```diff
@@ -12,10 +12,8 @@
     <% end %>
   </li>
 
-  <%= menu_separator() %>
-
   <li role="presentation" class="<%= 'active' if params[:starred].present? %>">
-    <%= link_to root_path(starred: true, archive: params[:archive]), class: 'filter' do %>
+    <%= link_to root_path(starred: true), class: 'filter' do %>
```
briannelson: You can remove `class: 'filter'` from here so that Starred is displayed the same.

app/views/notifications/_filter-list.html.erb
```diff
@@ -33,13 +33,6 @@
         <% end %>
       <% end %>
 
-      <% if filters[:starred].present? %>
-        <%= link_to root_path(filters.except(:starred)), class: 'btn btn-default' do %>
-          <%= octicon 'x', :height => 16 %>
-          <%= (params[:starred] === 'true'? 'Starred' : 'Unstarred') %>
-        <% end %>
-      <% end %>
-
```
knolleary: I'm out of sync here as to whether we want 'starred' to appear in the filter list or not (as you can tell, I added it back after you removed it). If the decision is to remove it, then the `if` statement at the top of this partial (line 3) will need to have `starred` removed from its clause so we don't show an empty `filter` (as can be seen in your screenshot).

andrew:
@tarebyte yeah I think this is what we want, we can tweak the behaviour further when the search bar comes in 👍

andrew:
🚢 to heroku

PR #118, opened 2016-12-22
add `o` shortcut for opening issues

zeke:
👋 

This PR adds `o` as a shortcut for opening issues in a new window, alongside `Enter`. This behavior is consistent with Gmail, Google Inbox, [ntl](http://ghub.io/ntl), and probably others. Vim? I dunno.


tarebyte:
Thanks @zeke!

PR #131, opened 2016-12-22
Ensure sidebar types with missing icons stay aligned

### LINKED ISSUE #129: 'Release' types in sidebar missing icon
The Release type in the sidebar has no icon, so appears misaligned:

cursor_and_octobox

Either need to add an icon (not sure if there is one...) or pad it out to keep things lined up.

knolleary:
fixes #129


andrew:
👌

PR #135, opened 2016-12-22
Setup caching for Octobox

tarebyte:
Closes https://github.com/andrew/octobox/issues/125

This adds middleware for caching so that we're not wasting our rate limit.

/cc @andrew for 👀


andrew:
Nice, will merge and deploy this a little later when I'm at my laptop 💻

PR #136, opened 2016-12-22
Make pagination a little more GitHubby

tarebyte:
This updates the pagination to look at little more like what the pagination looks like on GitHub.

screen shot 2016-12-22 at 16 55 06

screen shot 2016-12-22 at 17 31 39

I've also made it mobile friendly by changing the window size.

/cc @briannelson for 👀 

I'm also totally cool if this doesn't get merged just thought I'd see what people thought.


andrew:
👍

PR #137, opened 2016-12-22
Don't spin icon on page refresh

tarebyte:
When I refresh manually using CMD-R the spinner spins.

spin-on-refresh

This fixes it by making sure that the class is only added when the icon itself is clicked.

/cc @andrew


andrew:
😂 good catch!

andrew:
I sleepily tried to [revert](https://github.com/andrew/octobox/commit/803f58116b25bf8dab5de6b4b5cdc28701311e66) this this morning as I noticed the spinner didn't respond to `r` being pressed 🙈  

I wonder if we can check for `event.metaKey` to see if CMD (or CTRL) has been pressed along with R

PR #142, opened 2016-12-23
Ensure archived/starred selection is preserved when filter is removed

knolleary:
As identified here https://github.com/andrew/octobox/issues/134#issuecomment-268930716


andrew:
🚢

PR #143, opened 2016-12-23
Move some filter logic into helper methods

andrew:
Just gets some of the filter logic away from the html

cc @knolleary


knolleary:
LGTM

PR #150, opened 2016-12-23
Multiple select

### LINKED ISSUE #60: Allow archiving multiple notifications simultaneously
I'm envisioning pressing `x` to mark the checkbox for a notification, then `y` will archive all of the currently selected notifications.

### LINKED ISSUE #68: UX question
I did not find that a checkbox to archive was intuitive enough (especially considering that #60 has not yet landed so only one can be selected at a time)

Related (as this is my current blocker to making a pr) is there any way to work with just the client side and have it connect to octobox.io or similar?

karpiu:
fixes #60 
fixes #68

The js code has become quite big. Do you want me to write some tests for it? What framework/gem you want to use?


andrew:
@karpiu I've refactored the javascript quite a lot this morning, there's a few conflicts there to resolve, feel free to throw out any of my refactoring that gets in the way.

karpiu:
dont merge yet. I found a bug

andrew:
Ok!

Just had a play around with the branch, it's working great! I think we can swap the "Archive All" button for a select all checkbox at the top of the list, similar to Gmail.

karpiu:
@andrew I believe everything is ok now. Play around some more and tell me if you spot some weird cursor jumps after archiving (I may have missed some cases). New position of the cursor after archiving should be calculated like this:
- if cursor is on the unmarked item, then it should stay there after archive.
- if cursor is on the marked item, it finds the first item below that is unmarked and stays there after archive
- if cursor is on the marked item and all items below are also marked, then it finds the last unmarked item on the list and stays there after archive.
- if all items are marked, cursor should be at the top of list after archive (unless of course the list is empty after archiving, then there should be no cursor visible)

andrew:
Yeah I think those positions make sense 👍

andrew:
Just deployed :shipit:

PR #157, opened 2016-12-23
Use repo icon for visual aid on sidebar

tarebyte:
screen shot 2016-12-23 at 11 10 31

/cc @briannelson

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -220,7 +220,11 @@ td.keys {
       padding: 5px 15px;
 
       &.repo-label {
-        padding-left:30px;
+        padding-left: 25px;
+
+        .octicon {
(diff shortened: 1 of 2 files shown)


andrew:
I wonder if it makes sense to only group by org when there are more than one repo to show? Not sure how the mix of orgs and repos would work though, might be more confusing

tarebyte:
> I wonder if it makes sense to only group by org when there are more than one repo to show? Not sure how the mix of orgs and repos would work though, might be more confusing

I don't mind having it being grouped with only one, but that may just be me. I just thought it was weird to not have an icon.

briannelson:
With the indentation, this looks pretty slick. Nice work.

octobox

andrew:
Agreed it is weird without the icon

andrew:
🚢

PR #161, opened 2016-12-23
Replace reference to unused variable GITHUB_TOKEN

davidcelis:
In the past, a Personal Access Token was supplied to Octobox via the `GITHUB_TOKEN` variable. However, since Octobox is now implemented as an OAuth Application, we use a pair of `GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET` variables instead. The README still contains a reference to supplying `GITHUB_TOKEN` in a `.env` file, so let's fix it up to reference the real variables we use.

Signed-off-by: David Celis <me@davidcel.is>


andrew:
Good catch!

PR #162, opened 2016-12-23
Keep the full name of the repo for the filter

tarebyte:
repo-name

Thoughts?


davidcelis:
I like this; I'd prefer seeing the whole name-with-owner instead of just the repo name 😄

andrew:
Yeah that's better

Issue #52, opened 2016-12-19
Search


andrew:
Without opening a huge can of worms or adding any extra dependencies I wonder if we can add a basic search using a postgres full text search index: http://rachbelaid.com/postgres-full-text-search-is-good-enough/

m1guelpf:
![search-dfQwvk9YLzVmM](https://media2.giphy.com/media/dfQwvk9YLzVmM/giphy.gif)

andrew:
This gem might make managing search queries and combinable filters nicer: https://github.com/marcelocf/searrrch

> Allow you to receive a query string using the syntax similar to gmail's filters and parse it without going crazy.

> The following search string is supported:

    user_id: 123 user_name: omg user_id: 124 free text here
> The user can also put the value between quotes, in which case nested escaped quotes are supported.

andrew:
Fixed by @samjacobclift

PR #164, opened 2016-12-23
Feature/search

royka:
***Work In Progress***

**headline**
1) Postgres based full text search plus addition to the UI for it :), see this [issue](https://github.com/octobox/octobox/issues/52) for more details

**todo**

- [x] Better UI, see ticket for more details
- [x] Make sure other query params are not reset/removed by the filter
- [x] Make sure that shortcut commands don't get triggered when typing
- [x] Add gmail like filtering (to do later)
- [x] Add index (to do later)

**Screenshot**

Mobile
(only appears once dropdown has been selected)

screen shot 2016-12-30 at 18 39 00

Desktop
screen shot 2016-12-30 at 18 38 54

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 class Notification < ApplicationRecord
+  include PgSearch
+  pg_search_scope :search_by_subject_title, :against => :subject_title
+
   belongs_to :user
 
   scope :inbox,    -> { where(archived: false) }
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -16,7 +16,9 @@
     <div class="collapse navbar-collapse js-octobox-menu" id="octobox-menu">
       <%= form_tag root_path, method: :get, class: 'navbar-form navbar-left', enforce_utf8: false do |f| %>
         <div class="form-group">
-          <%= hidden_field_tag :parameters, request.query_string  %>
+          <% @text_search_hidden_field_params.each do |name, value| %>
+            <%= hidden_field_tag name, value  %>
+          <% end %>
           <%= text_field_tag 'q', "#{@query || '' }", placeholder: 'Search...', class: 'form-control', size: 40, id: 'search-box' %>
         </div>
       <% end %>
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -17,7 +17,10 @@
       <%= form_tag root_path, method: :get, class: 'navbar-form navbar-left', enforce_utf8: false do |f| %>
         <div class="form-group">
           <% @text_search_hidden_field_params.each do |name, value| %>
-            <%= hidden_field_tag name, value  %>
+            <!-- filter out unwanted query params -->
+            <% if (name != 'controller') && (name != 'action') %>
+              <%= hidden_field_tag name, value  %>
+            <% end %>
           <% end %>
           <%= text_field_tag 'q', "#{@query || '' }", placeholder: 'Search...', class: 'form-control', size: 40, id: 'search-box' %>
         </div>
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -16,6 +16,7 @@
     <div class="collapse navbar-collapse js-octobox-menu" id="octobox-menu">
       <%= form_tag root_path, method: :get, class: 'navbar-form navbar-left', enforce_utf8: false do |f| %>
         <div class="form-group">
+          <%= hidden_field_tag :parameters, request.query_string  %>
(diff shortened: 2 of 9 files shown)


royka:
Initial PR just for some feedback, I'm a bit of a rails noob :P

andrew:
Nice start! I was just thinking about looking into this 👌

andrew:
pinging @briannelson for his thoughts on UI

briannelson:
My initial thought is that the search bar should be either in the top navbar or directly below the top navbar, above the table and filters. 

Placing it in the header of the table suggests it filters only the notifications that are visible within the current visible table.

My other reason for placing it in its own row would be for improved (read: easier to implement) mobile performance. 

Something like this?

octobox

octobox

royka:
Nice, sounds good I'll work on making it look like that 👍

andrew:
@royka a few conflicts here that you might run into

royka:
@andrew yeah rebased and seems good 👍 also @briannelson wdyt of those screenshots? I can change it to your original ideas if you like

tarebyte:
I'm not a fan of the mobile just because we're hiding search in the navigation dropdown.

briannelson:
@royka 👍 I like it in the navbar!

briannelson:
@tarebyte maybe we expose it on mobile all the time? Might require two separate search inputs...

royka:
That search box layout is ripped straight out of bootstrap ;) I mean we could move the text input into the side bar filter options, that might work? Thing is if we always show it on mobile might take up a-lot of screen real estate

royka:
Also we might be better to merge this when we are happy and add the advanced search options in a later PR? I'm not too worried though if we want to do it in one go

app/views/layouts/_header.html.erb
```diff
@@ -14,6 +14,17 @@
 
     <!-- Collect the nav links, forms, and other content for toggling -->
     <div class="collapse navbar-collapse js-octobox-menu" id="octobox-menu">
+      <%= form_tag root_path, method: :get, class: 'navbar-form navbar-left', enforce_utf8: false do |f| %>
+        <div class="form-group">
+          <% @text_search_hidden_field_params.each do |name, value| %>
+            <!-- filter out unwanted query params -->
+            <% if (name != 'controller') && (name != 'action') %>
+              <%= hidden_field_tag name, value  %>
+            <% end %>
```
royka: so the logic here is to create a hidden field for every query param we currently have, that way when the form is submitted we keep the existing, filters

app/models/notification.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 class Notification < ApplicationRecord
+  include PgSearch
+  pg_search_scope :search_by_subject_title, :against => :subject_title
+
```
royka: Also this doesn't by default use a gin index 😬 we would have to create one manually (and with a migration) if we find that the search is too slow. @andrew wdyt

andrew: Yeah let's add an index in a migration 👍

tarebyte:
> maybe we expose it on mobile all the time? Might require two separate search inputs...

If other people like it in the nav that's totally fine, it's just a personal preference of mine.

briannelson:
I'd prefer not to have it in the sidebar, I think the top navbar is a more appropriate place. Let's leave as is for now and I can adjust the navbar input so that it shows up in the top navbar regardless of screen size.

andrew:
I like it in the header on desktop, would prefer it hidden behind an icon on mobile though

db/migrate/20161231115932_add_full_text_search_indexes.rb
```diff
@@ -0,0 +1,9 @@
+class AddFullTextSearchIndexes < ActiveRecord::Migration[5.0]
+  def up
+    execute "create index notifications_subject_title on notifications using gin(to_tsvector('english', subject_title))"
+  end
+
+  def down
+    execute "drop index notifications_subject_title"
+  end
+end
```
royka: so this is a gin index for the 'subject_title' field, I ran an explain analyse on a text search query however it didn't seem to use this index 😬 maybe its because I need to vacuum analyse or maybe its not efficient to use this index on my small local dataset. However if we wanted to increase the number of fields to be included in the text search we will need to create another field to hold tokens of all the fields we want to search across.

app/models/notification.rb
royka: added see the latest comment :)

briannelson:
@andrew 👍

db/migrate/20161231115932_add_full_text_search_indexes.rb
andrew: I tried it on a bigger db and it still wasn't used, but after filtering by user_id it's really quick at the moment, let's remove this for now and revisit again if performance get's bad once the db is much bigger!

royka: nice done

andrew:
Merging and shipping 🚢

PR #165, opened 2016-12-23
Scroll with cursor

### LINKED ISSUE #160: Let the cursor move the page up
When I use the ✨ shortcuts if I scroll too far up or too far down the page doesn't move.

moving-page

karpiu:
Fixes #160 

It was surprisingly tricky to make it work both on Firefox and Chrome. Please check other browsers before merging.

Also I had to remove `height: 100%` from `html` and `body`, otherwise Chrome was not cooperating. Not sure what kind of consequences this has on the rest of the app. Someone please shed some light on this issue.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -7,7 +7,7 @@
 document.addEventListener("turbolinks:load", function() {
   $('button.archive_selected, button.unarchive_selected').click(function () { toggleArchive(); });
   $('input.archive, input.unarchive').change(function() {
-    marked = $(".table-notifications input:checked");
+    marked = $(".js-table-notifications input:checked");
     if ( marked.length > 0 ) {
       $('button.archive_selected, button.unarchive_selected').show();
     } else {
@@ -25,7 +25,7 @@ document.addEventListener("turbolinks:load", function() {
 });
 
 document.addEventListener("turbolinks:before-cache", function() {
-  $('td.current').removeClass("current");
+  $('td.js-current').removeClass("current js-current");
 });
 
 // Add shortcut events only once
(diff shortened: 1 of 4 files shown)


andrew:
Would it make it easier if the navbar was static rather than sticky?

karpiu:
@andrew It would not make much difference code-wise.

app/assets/javascripts/application.js
```diff
@@ -131,6 +131,26 @@ function sync() {
...
 }
 
+function scrollToCursor() {
+  table_offset = $('.table-notifications').position().top;
+  cursor_offset = $('td.current').offset().top;
+  cursor_relative_offset = $('td.current').position().top;
+  cursor_height = $('td.current').height();
+  menu_height = $("#octobox-menu").height();
+  scroll_top = $(document).scrollTop();
+  window_height = $(window).height();
```
tarebyte: Would you mind using `js-` prefixed CSS classes? That way if something changes we know whether it's for styling or for functionality.

My apologies if I've reviewed something and haven't said anything before.

karpiu: So you want me to change `current` to `js-current`?

tarebyte: Yeah so the markup would have things like `js-table-notifications` `js-current` and `js-octobox-menu`

karpiu: And the old classes should stay the same? You want to have both `js-table-notifications` and `table-notifications` on this table? Because there are styles for `table-notifications` in `application.scss` already.

tarebyte: Correct!

briannelson:
> Also I had to remove height: 100% from html and body

@karpiu This was originally added to keep the filters visible on mobile screens when the notifications table was shorter than the viewport. Because we have min-height: 100%, I think we're still good.

karpiu:
@tarebyte @andrew  ready for another review

andrew:
LGTM!

PR #167, opened 2016-12-24
Add icons for dock usage

briannelson:
If you’re using an app like Fluid, these icons should be a little nicer.

octobox


andrew:
You read my mind 😍

PR #168, opened 2016-12-24
refactor archive/unarchive selected button

karpiu:


app/helpers/notifications_helper.rb
```diff
@@ -23,4 +23,12 @@ def any_active_filters?
   def filtered_params(override = {})
     filters.merge(override)
   end
+
+  def archive_selected_button(custom_class=nil)
+    action = params[:archive] ? 'unarchive' : 'archive'
+    button_tag(type: "button",
+               class: "archive_toggle #{action}_selected #{custom_class}") do
+      "#{action} selected".capitalize
```
michaelrauh: Would it make sense to remove the `.capitalize` and replace 
`action = params[:archive] ? 'unarchive' : 'archive'`
with 
`action = params[:archive] ? 'Unarchive' : 'Archive'`?
I think it would be the same, but correct me if I'm wrong.

andrew: 👍

karpiu: @michaelrauh Variable `action` is used in the line 30 to construct a class name `archive_selected` or `unarchive_selected`. So the question is: are HTML class names case-sensitive?

michaelrauh: @karpiu Good call out! From what I have read, html is 'mostly' case-sensitive. I gave my suggested change a try, and it breaks things. Sorry about that!

PR #169, opened 2016-12-24
Improve Docker development environment

bfirsh:
A follow-up from #35.

- Wait for database to start on boot so only a single command is needed.
- Tidied up readme.
- Explicitly specify build steps instead of using onbuild so the apt package cache isn't busted by new code.
- Rename built image to be "octobox"

This is ridiculously simpler than installing manually. Perhaps it makes sense to be the recommended way to get started, with the manual steps as the alternative if you don't want to use Docker? 😄

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,16 @@
-FROM ruby:2.3-onbuild
-
+FROM ruby:2.3
 RUN \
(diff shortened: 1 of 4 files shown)


andrew:
That's wonderful, thanks @bfirsh!

PR #170, opened 2016-12-25
Removed deploy to heroku

GiacomoLaw:
Not really needed as there is a website for it.


andrew:
Quite a few people run this app themselves as they don't want to give me the scope to read their notifications, deploying to heroku is an easy and cheap way to get up and running so I'm not going to remove this

GiacomoLaw:
Okay, cool! Sorry about that.

PR #171, opened 2016-12-25
Set page to last if page number is too big

### LINKED ISSUE #166: Archiving all notifications on last page incorrectly tells you your octobox is open
#### To replicate:
1. Have at least 2 pages worth of notifications.
2. Go to last page.
3. Archive all notifications on that page.

#### What should happen:
Octobox should show the previous page (the new last page)

#### What happens:
Octobox shows a notification telling you: You're in the clear! There are no notifications that need your attention.

#### Easier way to replicate:
1. Visit https://octobox.io/?page=20000

_If this page still shows notifications... you have bigger problems than this bug to deal with._

karpiu:
Fixes #166

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -26,7 +26,9 @@ class NotificationsController < ApplicationController
     scope = scope.status(params[:status]) if params[:status].present?
     scope = scope.owner(params[:owner])   if params[:owner].present?
 
-    @notifications = scope.newest.page(page).per(per_page)
+    total_pages = (scope.count / per_page.to_f).ceil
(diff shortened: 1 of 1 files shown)


andrew:
Can we redirect to the correct URL instead of changing the page number in the action?

karpiu:
@andrew But that would mean recalculating everything in `index` action, right? Would it matter that much, performance-wise?

karpiu:
@andrew Have you read my previous comment?

andrew:
@karpiu just committed a tweaked version of this that redirects if the page is out of bounds: https://github.com/octobox/octobox/commit/544238b19077871d4fb8b6730abeef50d3042e0e

PR #174, opened 2016-12-27
Add user id to notifications unique index

andrew:
Turns out that the `id` in the [notifications api](https://developer.github.com/v3/activity/notifications/) is not unique for users, so multiple users actually get the same notification record for a single pull request sent to all the users watching the repo.

We assumed it was unique and so added a unique database index to make sure we don't get any duplicates but this turns out to cause problems!

Anyone running their own instance wouldn't run into this but the hosted version has multiple people signed up and watching the same repos, so who ever gets the notification first will block everyone else from seeing it.

This pr changes the index to be unique by user id and notification id 😅


andrew:
Just deployed this to octobox and found an extra 54 notifications on my account that must have been blocked by the previous index 🙈

PR #178, opened 2016-12-27
README: Update "you have to host it yourself" to reflect SaaS progress.

apetro:


andrew:
Ah yeah, good catch, completely forgot about that!

PR #180, opened 2016-12-27
Ruby 2.4.0

andrew:
Release notes: https://www.ruby-lang.org/en/news/2016/12/25/ruby-2-4-0-released/

Not sure if we should upgrade yet so opening a pr to discuss.

Tests pass and the app works fine but some people may have problems getting the latest version and there's no ruby 2.4.0 on docker hub yet: https://hub.docker.com/r/library/ruby/tags/ so might be worth waiting a few weeks until its more widely available?


tarebyte:
I'd hang on for like a week just to be on the safe side.

WillAbides:
You should change the ruby version in Dockerfile too

andrew:
@WillAbides just waiting for ruby 2.4.0 final to be available: https://hub.docker.com/_/ruby/

WillAbides:
Doh.  I didn't read the comment.

andrew:
Ruby 2.4 docker image is available now, updated PR to use it

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.3
+FROM ruby:2.4
```
WillAbides: I would rather this be ruby:2.4.0.  I had trouble with the existing Dockerfile because my cached ruby:2.3 was 2.3.1 which didn't match the ruby specified in Gemfile.  We won't have that problem if we use same granularity in Gemfile and Dockerfile.

andrew:
@WillAbides I didn't realise that was a thing, have updated the docker file to specify 2.4.0 and will update master to 2.3.3 for now too.

andrew:
🚢

PR #181, opened 2016-12-27
Collapse sub_scope logic into one looping conditional

michaelrauh:
This is a small refactor to lower the complexity of the scope selecting logic in the notifications controller index function. It loops over the sub_scopes instead of handling them separately. A similar refactor was tried on the complexity_refactor PR but it did not update scope properly, and was more complex as it had to take starred into account (starred has been removed from this section of the logic).


andrew:
👌 very nice!

PR #182, opened 2016-12-28
Use label for select_all button

knolleary:
This replaces the select_all button `div` with a `label` that has its `for` attribute set. This allows us to make the whole button clickable, rather than just the select box inside it.


andrew:
Good thinking!

Issue #44, opened 2016-12-19
Allow option to authenticate with a personal access token


WillAbides:
Add the option to authenticate with a personal access token.  This was lost in #25.

jlmitch5:
As far as I can tell, there's no way to provide the "notifications scope" mentioned in the readme.md anymore with the deploy to heroku button now, as it doesn't seem like there's a way to set that in the "Register a New oauth application" screen. This may not matter, but just wanted to write a note in case it does.  These are the config vars you get in heroku now:

<img width="962" alt="screen shot 2016-12-19 at 12 51 17 pm" src="https://cloud.githubusercontent.com/assets/1342624/21323126/1d786bfc-c5ea-11e6-9059-32f5c43aef90.png">

Readme.md should also probably be updated to explain the new config vars.

andrew:
@jlmitch5 ah good catch, that's definitely got a bit out of sync today, the app now defaults to `notifications` as the scope so you don't need to set it, and the readme is definitely out of date around the heroku button

jlmitch5:
Sounds good!  Would pr a change to the readme, but it sounds like there's probably active development/changes around adding the personal access token config back in, so I thought it best to hold off until that lands.

andrew:
@jlmitch5 I think the PR will add the ability into the UI rather than via ENV, will happily merge a quick update to the readme now

WillAbides:
I'm going to have a go at this one.  @tarebyte, do you have any words of wisdom since you touched this part recently?

tarebyte:
@WillAbides shouldn't be too bad honestly. I [started a branch](https://github.com/octobox/octobox/tree/tarebyte/edit-account) but got sidetracked.

The only thing I would make sure to do is to make sure the users token doesn't get overridden once they do the oauth dance a second time.

andrew:
Maybe store it in a separate column?

andrew:
Fixed by @WillAbides

Issue #175, opened 2016-12-27
Allow users to delete their account


andrew:
Would be good to make deleting your account a little more self service, perhaps from some kind of user settings page?

PR #185, opened 2016-12-29
Optionally use personal access token

WillAbides:
This implements #44.

There are a few things left to do
- [x] Show field errors on the user-preferences form
- [x] Work out a way to submit user-preferences without changing the token.  That isn't that important now, but we will probably have more preferences in the future.
- [x] Don't show personal access token on the user-preferences form if that is disabled
- [x] Add text to the user-preferences form explaining what a personal access token is and how to create one.
- [x] Add an environment variable to enable the use of personal access tokens.

I want to call out the last one especially.  It isn't mentioned in #44, but I think we should have this feature disabled by default.  Because these tokens are equivalent to GitHub passwords and we are storing them in plain text, it wouldn't be responsible to store them in a shared environment.


andrew:
> Add an environment variable to enable the use of personal access tokens.

Agreed, making that opt in is a great idea 👌

andrew:
@WillAbides we should also add some documentation around this feature in the readme as it's more aimed at people running their own instance than for the hosted version.

test/controllers/users_controller_test.rb
```diff
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'pp'
```
andrew: we can remove this right?

WillAbides: Oops.  I'll remove it in my next commit.

WillAbides:
I've gone through my todo list and added to readme.  Removing [WIP] from the title, and this is for review.

andrew:
Looking good @WillAbides, just need a `ensure_logged_in` before_action on the users controller to make sure the user is logged in

tarebyte:
> just need a `ensure_logged_in` before_action on the users controller to make sure the user is logged in

What about adding the `before_action` to the `ApplicationController` and then skipping it in the `SessionsController`? That way we default to being logged in.

andrew:
@tarebyte currently the notifications controller also renders the homepage 🙈

WillAbides:
I ended up using the `before_action :authenticate_user!` that is inherited from ApplicationController and adding `ensure_correct_user` to make sure you can't update a different user.

andrew:
LGTM, any objections @tarebyte?

WillAbides:
@andrew: You said redirect to /login before my last push.  I guess you are ok with it redirecting to / instead?

andrew:
@WillAbides oh yeah that's fine, the session controller would forget where they came from anyway!

app/models/user.rb
```diff
@@ -16,7 +23,7 @@ def assign_from_auth_hash(auth_hash)
     github_attributes = {
       github_id: auth_hash['uid'],
       github_login: auth_hash['info']['nickname'],
-      access_token: auth_hash.dig('credentials', 'token')
+      access_token: auth_hash.dig('credentials', 'token'),
```
tarebyte: I know some people prefer to have to have a trailing commas to make the diff cleaner, but my inner rubocop is sad.

What do you think about undoing this change?

config/initializers/octobox_config.rb
```diff
@@ -0,0 +1,12 @@
+module Octobox
```
tarebyte: Does this need to load before any other configs?

If so, might be helpful to prefix this with a `02`

app/models/user.rb
WillAbides: It's gone.  This was purely unintentional.

config/initializers/octobox_config.rb
WillAbides: The order shouldn't matter here.

config/routes.rb
```diff
@@ -20,4 +20,7 @@
       get :star
     end
   end
+
+  get '/user_preferences', to: 'users#edit'
```
tarebyte: Not a huge fan of `user_preferences` as a url. What do you think about changing it to `settings`?

tarebyte:
Left a few comments, looks fine to me @andrew

config/routes.rb
WillAbides: Settings is fine with me.  My main reason for changing it from /edit was to avoid ambiguity about what was being edited.

andrew:
Cool! We can do https://github.com/octobox/octobox/issues/175 nicely once this is merged

WillAbides:
I was already thinking about doing #175 next.

andrew:
:shipit:

bronislav:
@andrew Is this feature available at octobox.io ?

andrew:
@bronislav no it's not at the moment, I didn't feel comfortable storing people's personal access tokens plaintext the heroku database.

bronislav:
@andrew Thank you for clarification

andrew:
@bronislav I've opened #212 to get those features documented in the readme

PR #189, opened 2016-12-30
Allow users to self-delete

### LINKED ISSUE #175: Allow users to delete their account

WillAbides:
closes #175 

This adds a button to /settings that will delete a user's account.

2016-12-30 at 1 57 pm


andrew:
👍

PR #192, opened 2016-12-31
Add memcached config to docker-compose.yml

andrew:
I think this is correct, not done much docker composing before though so would be good to get a second opinion

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,8 @@ services:
     image: postgres:9.6
     environment:
       - POSTGRES_PASSWORD=development
-
+  memcached:
+    image: memcached
   app:
     image: octobox
     build:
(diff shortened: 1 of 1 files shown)


tarebyte:
@andrew won't we have to update the `config/environments/development.rb` file in order for this to be used?

andrew:
D'oh! Yeah good point @tarebyte, I forget that the docker is only for development atm

docker-compose.yml
```diff
@@ -23,4 +25,5 @@ services:
       - OCTOBOX_DATABASE_USER=postgres
       - OCTOBOX_DATABASE_PASSWORD=development
       - OCTOBOX_DATABASE_HOST=database.service.octobox.internal
+      - MEMCACHEDCLOUD_SERVERS=memcached.service.octobox.internal
```
tarebyte: If we want to mirror a "production" environment in development, we may also need to set the username and password for the memcached service ([file](https://github.com/octobox/octobox/blob/master/config/environments/production.rb#L56-L59))

andrew: Agreed, would be good to make the docker setup usable for dev, test and production

andrew: Turns out standard memcached doesn't have any username/password support built in, memcachedcloud adds that on top of it.

PR #195, opened 2016-12-31
Sync notifications on login

### LINKED ISSUE #193: Suggestion: Sync notifications on every new login
After logging in, it would be great to have the notifications synced to make it more convenient. Syncing on every load would be too much, but syncing on every login seems very reasonable to me.

tarebyte:
Closes #193 

/cc @andrew

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -11,6 +11,7 @@ class SessionsController < ApplicationController
     user      = User.find_by_auth_hash(auth_hash) || User.new
 
     user.assign_from_auth_hash(auth_hash)
+    user.sync_notifications
 
     session[:user_id] = user.id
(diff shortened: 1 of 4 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -11,6 +11,7 @@ def create
     user      = User.find_by_auth_hash(auth_hash) || User.new
 
     user.assign_from_auth_hash(auth_hash)
+    user.sync_notifications
```
andrew: If we're doing it in the controller we can remove the after create callback here: https://github.com/octobox/octobox/blob/master/app/models/user.rb#L10

tarebyte: Resolved in 45dbfda

andrew:
🚢

Issue #199, opened 2016-12-31
Filters are lost when syncing notifications


WillAbides:
When you are viewing results filtered by reason or archive status then refresh notifications, your filters are lost.

This was also discussed here:  https://github.com/octobox/octobox/issues/198#issuecomment-269874254

PR #200, opened 2016-12-31
Auto refresh notifications page

### LINKED ISSUE #198: Allow user to automatically sync periodically
I'd like to be able to keep octobox in the background and glance at it periodically to see if anything new is on top.

If I could schedule the sync to every n minutes, I wouldn't have to change window focus and hit the refresh button.

WillAbides:
closes #198
no longer blocked by #199

This refreshes notifications on an interval set by the user

### DIFF
--- a/app.json
+++ b/app.json
@@ -13,6 +13,10 @@
     "GITHUB_CLIENT_SECRET": {
       "description": "The GitHub Application Client Secret",
       "required": true
+    },
+    "MINIMUM_REFRESH_INTERVAL": {
+      "description": "Minimum value allowed for a user to set auto-refresh.  Setting to 0 means this is disabled.",
+      "value": "10"
     }
   },
   "addons": [
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -159,14 +159,16 @@ function autoSync() {
   }
 }
 
-function setAutoSyncTimer(interval) {
-  if (isNaN(interval)) {
+function setAutoSyncTimer() {
+  if (isNaN(refresh_interval)) {
     return;
   }
-  if (interval > 0) {
-    setInterval(autoSync, interval)
+  if (refresh_interval > 0) {
+    setInterval(autoSync, refresh_interval)
   }
 }
+
+$(document).ready(setAutoSyncTimer);
 
 function scrollToCursor() {
   var current = $('td.js-current');
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -152,6 +152,22 @@ function sync() {
   $("a.sync").click();
 }
 
+function autoSync() {
+  marked = $(".js-table-notifications input:checked")
+  if ( marked.length == 0 ) {
+    sync()
+  }
+}
+
+function setAutoSyncTimer(interval) {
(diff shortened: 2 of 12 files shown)


WillAbides:
I opted for the settings page to set the refresh interval in one minute increments so that people don't end up setting something ridiculously short.

db/migrate/20170101214034_add_default_value_to_users_refresh_interval.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultValueToUsersRefreshInterval < ActiveRecord::Migration[5.0]
+  def change
+    change_column_null :users, :refresh_interval, false, 0
```
WillAbides: I'm not sure why this didn't set the default value to 0

tarebyte: I'm pretty sure it's `default: 0`

WillAbides: Looks like I had the syntax right, but it didn't do what I thought.  That just changes existing rows to 0.  [The documentation](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_null) even says "Please note the fourth argument does not set a column's default."

I just needed to RTFM

WillAbides:
I cleaned this up and force pushed so the commit log isn't so messy

WillAbides:
@andrew: Given your concerns in #205 about hitting the api too hard, maybe I should add a system config to enable this and set the minimum refresh time.

Something like `MINIMUM_REFRESH_INTERVAL`.  If it is unset or 0, this is disabled.  Otherwise it is the minimum number of minutes a user is allowed to set refresh_interval to.

andrew:
@WillAbides makes sense, I'm not going to be using this feature or #205 as I like to hit refresh once I'm done with everything I've already got in my inbox so would be good to get feedback from other users too

WillAbides:
I've now put this behind a system config.  The default minimum refresh interval is 10 minutes.

I think this is ready for merging.

andrew:
@WillAbides just need to change the `javascript_tag` to be a bit more unobtrusive, i.e. move the function to be called after document.ready in application.js and check for the presence of a data attribute in the page with the value of the interval.

WillAbides:
@andrew: Is https://github.com/octobox/octobox/pull/200/commits/0fdedf0d677e649aebe6b4ed60e214ead820ce8b what you're looking for?

andrew:
@WillAbides Nearly there but rather than setting a js variable in the html, add `data-refresh-interval=<%= current_user.effective_refresh_interval %>` to the main `<table>` on the page, then load it in your `setAutoSyncTimer` function with `$('.js-table-notifications').data('refresh-interval')`

WillAbides:
@andrew: Maybe this is better

andrew:
👍

andrew:
FYI I've just enabled auto-refresh on octobox.io with a 5 minute minimum: https://github.com/octobox/octobox/commit/f9b640ea5d0850dbe98cab6bf3cd2071d109d158

Issue #125, opened 2016-12-22
Better notification syncing with "If-Modified-Since" header


andrew:
If a notification has been marked as read on GitHub, but hasn't changed state and you don't have any new notifications then the current syncing method won't download the change in unread state.

That's because we pass `user.last_synced_at` as the `:since` option when downloading notifications, which is a little bit annoying as there's no way for octobox to mark the notifications as read if there's no new activity on GitHub.

Fortunately there's another way to get round this listed in the Notifications API docs: https://developer.github.com/v3/activity/notifications/

> Notifications are optimized for polling with the "Last-Modified" header. If there are no new notifications, you will see a "304 Not Modified" response, leaving your current rate limit untouched. There is an "X-Poll-Interval" header that specifies how often (in seconds) you are allowed to poll. In times of high server load, the time may increase. Please obey the header.

```
# Add authentication to your requests
curl -I https://api.github.com/notifications
HTTP/1.1 200 OK
Last-Modified: Thu, 25 Oct 2012 15:16:27 GMT
X-Poll-Interval: 60

# Pass the Last-Modified header exactly
curl -I https://api.github.com/notifications
   -H "If-Modified-Since: Thu, 25 Oct 2012 15:16:27 GMT"
HTTP/1.1 304 Not Modified
X-Poll-Interval: 60
```

I don't think that Octokit supports this out of the box so we have two options:

1. Use some custom middleware with octokit like I do in Libraries.io with caching and 304 support: https://github.com/librariesio/libraries.io/blob/master/config/initializers/octokit.rb

2. Don't use Octokit to download the notifications and instead do a plain request including the `If-Modified-Since` header with the value of `user.last_synced_at` and always send a `since` value of `1.month.ago.iso8601`

tarebyte:
> Use some custom middleware with octokit like I do in Libraries.io with caching and 304 support: https://github.com/librariesio/libraries.io/blob/master/config/initializers/octokit.rb

I'm very much in favor of using Octokit and adding caching.

tarebyte:
@andrew do we want to add memcached for a caching layer?

andrew:
@tarebyte yeah memcached sounds good to me, and if we use `Rails.cache` everywhere it allows other users to swap out the cache mechanism if they wish.

andrew:
Just shipped to heroku, had to make a couple tweaks and disable the connection pool option as it was exploding 🚢

PR #206, opened 2017-01-02
Implement refresh on refresh

### LINKED ISSUE #205: Refresh on refresh
The first thing I do every time I load https://octobox.io/ is to click the refresh button. Open to automating that `onload`?

WillAbides:
This gives the user an option to sync notifications every time the notifications page loads.

2017-01-02 at 1 41 pm

closes #205

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -5,6 +5,8 @@ class NotificationsController < ApplicationController
   before_action :find_notification, only: [:archive, :unarchive, :star]
 
   def index
+    current_user.sync_notifications if current_user.sync_on_load
+
     scope    = current_user.notifications
 
     scope = if params[:starred].present?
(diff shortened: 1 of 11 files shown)


andrew:
I think this is a bit too heavy handed, every time you click on any link in the sidebar it's going to sync, will make the app feel incredibly slow when it's enabled right?

WillAbides:
Yes it is a bit heavy-handed.  As was mentioned in #205, there's not a good, clean way to distinguish which loads should trigger a refresh.

One thing I considered is for sync to check last_synced_at and not trigger a new sync unless it has been at least a minute (or maybe more).  That still isn't ideal, but at least it keeps us from rapidly hitting the api when switching between filters.

WillAbides:
@andrew:  Consider addbec03c7227c3415778b48b1004f9ec0635bad.  Now when you sync notifications through auto-load, it sets since=last_synced_at to speed things up.  As you noted in #125, this won't sync read status.  For that you will still need to click the spinner.

andrew:
Closing this, reasoning here: https://github.com/octobox/octobox/issues/205#issuecomment-270030067

PR #207, opened 2017-01-02
Click a row to move the cursor

WillAbides:
I like the keyboard navigation, but I don't like having to start out by hitting `j` repeatedly to get to the first notification I want to mark.

This will let me click a row to set the cursor there.


andrew:
👍

PR #219, opened 2017-01-04
Fix scrollbars showing all the time

ifvictr:
With this commit, scrollbars will no longer show even if there is no content overflow. The scrollbar will now show when necessary.


andrew:
Thanks!

PR #225, opened 2017-01-06
Cache notification partial fragements

tarebyte:
I was looking at [Rails fragment caching](http://guides.rubyonrails.org/caching_with_rails.html#fragment-caching) and thought it might be nice to cache the views for Notifications since they don't change that often.

I noticed a bit of a speed up on my machine, but if it's an over optimization or has unintended side effects I'm 🆒 with this not being merged.

### DIFF
--- a/app/views/notifications/index.html.erb
+++ b/app/views/notifications/index.html.erb
@@ -42,7 +42,7 @@
 
         <% if @notifications.any? %>
           <table class='table table-hover table-notifications js-table-notifications' data-refresh-interval=<%= current_user.effective_refresh_interval %>>
-            <%= render partial: 'notification', collection: @notifications %>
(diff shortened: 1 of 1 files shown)


tarebyte:
Ok so I found a slight issue with this.

screen shot 2017-01-06 at 00 16 16

Because we don't update the timestamp when we star a notification we can't bust the cache to update this view.

I don't think we want to update the record when we star the notification, so this might not be worth it.

andrew:
I think this is overkill tbh, as you found out, we only let the github api update the update_at timestamp, so we could track all the changes ourselves in a different timestamp but we're loading all the notifications from the database anyway, the performance improvement would be very small in production compared to the extra code to manage cache expiry

PR #227, opened 2017-01-08
Make separate requests for read and unread notifications

### LINKED ISSUE #221: Rethinking stopgap fix for syncing old notifications
I'd like to rethink #127.  A simpler solution may be to have two different requests on sync.

The first request would get only unread notifications since last sync with parameters like `{ all: false, since: user.last_synced_at.iso8601 }`.

The second request would be the same as the current request for all notification, but its results would only be used to sync the read status.  The one change I would make to this request would be to make it request since user's oldest unread notification.

I think this change would have a couple of benefits.  First, it would reduce complexity that leads to issues like #215.  Second, it will lead to faster syncs when the user only has very recent unread items (or none at all).

We could also make the syncs appear much faster if we make the second request a background task with the tradeoff being that items aren't marked read until the next refresh.

WillAbides:
This decreases complexity by separating out api requests for read and unread notifications.

The request for unread notifications goes back to using the If-Modified-Since header because it will cause an `HTTP 304` response if there is nothing new and the rate limit won't be decremented.  

The request for read notifications now sets the since value to your oldest unread notification.  This will prevent loading more pages of notifications than is necessary to mark a notification as read.

closes #221

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -45,9 +45,13 @@ class Notification < ApplicationRecord
     def download(user)
       timestamp = Time.current
 
-      fetch_unread_notifications(user)
-      fetch_read_notifications(user)
+      if user.last_synced_at
+        fetch_read_notifications(user)
+      else
+        new_user_fetch(user)
+      end
 
+      fetch_unread_notifications(user)
       user.update_column(:last_synced_at, timestamp)
     end
 
@@ -79,7 +83,7 @@ class Notification < ApplicationRecord
       return if notifications.blank?
       notifications.each do |notification|
         next if notification.unread
-        n = user.notifications.find_by(github_id: notification.id)
+        n = user.notifications.find_or_initialize_by(github_id: notification.id)
         next unless n
         n.update_from_api_response(notification)
       end
@@ -100,6 +104,12 @@ class Notification < ApplicationRecord
         notifications = user.github_client.notifications(all: true, since: since.iso8601, headers: headers)
         process_read_notifications(notifications, user)
       end
+    end
+
+    def new_user_fetch(user)
+      headers = {cache_control: %w(no-store no-cache)}
+      notifications = user.github_client.notifications(all: true, headers: headers)
+      process_read_notifications(notifications, user)
     end
   end
 
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -26,58 +26,79 @@ class Notification < ApplicationRecord
 
   paginates_per 20
 
+  API_ATTRIBUTE_MAP = {
+    repository_id: [:repository, :id],
+    repository_full_name: [:repository, :full_name],
+    repository_owner_name: [:repository, :owner, :login],
+    subject_title: [:subject, :title],
+    subject_type: [:subject, :type],
+    subject_url: [:subject, :url],
+    reason: [:reason],
+    unread: [:unread],
+    updated_at: [:updated_at],
+    last_read_at: [:last_read_at],
+    url: [:url],
+    github_id: [:id]
+  }.freeze
+
   class << self
     def download(user)
       timestamp = Time.current
 
-      fetch_notifications(user) do |notification|
-        begin
-          n = user.notifications.find_or_initialize_by(github_id: notification.id)
-
-          if n.archived && n.updated_at < notification.updated_at
-            n.archived = false
-          end
-
-          attrs = {}.tap do |attr|
-            attr[:repository_id]         = notification.repository.id
-            attr[:repository_full_name]  = notification.repository.full_name
-            attr[:repository_owner_name] = notification.repository.owner.login
-            attr[:subject_title]         = notification.subject.title
-            attr[:subject_type]          = notification.subject.type
-            attr[:reason]                = notification.reason
-            attr[:unread]                = notification.unread
-            attr[:updated_at]            = notification.updated_at
-            attr[:last_read_at]          = notification.last_read_at
-            attr[:url]                   = notification.url
-
-            attr[:subject_url] = if notification.subject.type == "RepositoryInvitation"
-                                   "#{notification.repository.html_url}/invitations"
-                                 else
-                                   notification.subject.url
-                                 end
-          end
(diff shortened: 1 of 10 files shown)


app/models/notification.rb
```diff
@@ -26,58 +26,79 @@ class Notification < ApplicationRecord
...
+    subject_title: [:subject, :title],
+    subject_type: [:subject, :type],
+    subject_url: [:subject, :url],
+    reason: [:reason],
+    unread: [:unread],
+    updated_at: [:updated_at],
+    last_read_at: [:last_read_at],
+    url: [:url],
+    github_id: [:id]
+  }.freeze
```
WillAbides: I thought having a single map would be more maintainable than coding the paths in the update methods

app/models/notification.rb
```diff
@@ -26,58 +26,79 @@ class Notification < ApplicationRecord
...
       timestamp = Time.current
 
-      fetch_notifications(user) do |notification|
+      fetch_unread_notifications(user)
+      fetch_read_notifications(user)
+
+      user.update_column(:last_synced_at, timestamp)
+    end
+
+    def attributes_from_api_response(api_response)
```
WillAbides: I extracted this method primarily to be able to test it separately from the update method

app/models/notification.rb
```diff
@@ -26,58 +26,79 @@ class Notification < ApplicationRecord
...
+      attrs = API_ATTRIBUTE_MAP.map{|attr, path|
```
tarebyte: Would you mind using a `do` block here?

```ruby
attrs = API_ATTRIBUTE_MAP.map do |attr, path|
  [attr, api_response.to_h.dig(*path)]
end.to_h
```

WillAbides: Don't mind if I `do`

app/models/notification.rb
```diff
@@ -26,58 +26,79 @@ class Notification < ApplicationRecord
...
+      end.to_h
+      if "RepositoryInvitation" == api_response.subject.type
+        attrs[:subject_url] = "#{api_response.repository.html_url}/invitations"
+      end
+      attrs
+    end
+
+    private
+
+    def process_unread_notifications(notifications, user)
```
WillAbides: I extracted this and `process_read_notifications` with the idea of testing them separately.  I ended up not going that way with testing but left them separate.

test/fixtures/users.yml
```diff
@@ -10,3 +10,13 @@ tokenuser:
   access_token: <%= SecureRandom.hex(20) %>
   github_login: 'tokenuser'
   personal_access_token: 'deadbeef'
+
+userwithunread:
+  github_id: 92873
+  access_token: <%= SecureRandom.hex(20) %>
+  github_login: 'userwithunread'
+
+morty:
```
WillAbides: I added some new characters because I didn't want to break anybody's mental model of the existing test fixtures.

app/models/notification.rb
```diff
@@ -105,4 +126,20 @@ def web_url
...
+    if self.archived && change[1] > change[0]
+      self.archived = false
+    end
+  end
+
+  def update_from_api_response(api_response, unarchive: false)
+    attrs = Notification.attributes_from_api_response(api_response)
+    self.attributes = attrs
+    unarchive_if_updated if unarchive
+    save(touch: false) if changed?
```
WillAbides: It would be nice to have a separate `github_updated_at` field so that we don't end up accidentally tainting `updated_at` by saving without `touch: false`

test/models/notification_test.rb
```diff
@@ -64,4 +96,88 @@ class NotificationTest < ActiveSupport::TestCase
...
+      notification.attributes = Notification.attributes_from_api_response(n)
+      attrs = notification.attributes
+      notification.destroy
+      attrs.slice(*(keys.map(&:to_s)))
+    }
+  end
+
+  def notifications_from_fixture(fixture_file)
+    JSON.parse(file_fixture(fixture_file).read, object_class: OpenStruct).tap do |notifications|
+      notifications.map { |n| n.last_read_at = Time.parse(n.last_read_at).to_s if n.last_read_at }
```
WillAbides: We could simplify this by changing `last_read_at` from string to timestamp

andrew:
This is sweet!

test/models/notification_test.rb
andrew: 👍

app/models/notification.rb
andrew: 👍  then we could bring back @tarebyte's caching pr

test/fixtures/users.yml
andrew: 👍

app/models/notification.rb

andrew:
Looking good to me, only one thing I can think of that might be an issue, the first time a user logs in do we end up downloading all of their notifications twice?

WillAbides:
It won't double download on initial login.  However that reminds me of a behavior change that I should have mentioned earlier.

Currently, when you first login you get a month's worth of read notifications.  With this change, it only downloads unread notifications on initial login.  My thinking was that the initial month of reads was a side-effect of how the since date was handled and not really an intentional feature.

If you want to restore the initial month, it will be a pretty simple change.

andrew:
@WillAbides it should definitely download read messages for new users, the main problem that octobox solves is that github notifications are way too easily marked as read and disappear from the UI

WillAbides:
Indeed that is my favorite part of octobox.  I was thinking of it as a "from this point on" sort of thing though.  

I'll update it to get a month of read notifications on initial login.  Or maybe we should have it grab all available notifications instead.

andrew:
If we're only doing it once I think we can pull all their notifications

WillAbides:
I changed it so that new users get all notifications the api is willing to serve.  

On my account that is currently 502 notifications going back to 2017-01-01T09:07:07Z. My cursory search didn't find documentation of how far back it should go.

andrew:
Sweet, let's 🚢  and see how it goes!

PR #229, opened 2017-01-09
Remove memcached dependency

andrew:
Now that https://github.com/octobox/octobox/pull/227 has been merged I don't believe we are actually using memcached at the moment, this PR removes it and the custom octokit config that used it.


WillAbides:
I thought about taking it out in #227, but decided to leave it in case you had future plans for it.

I could see it coming in handy on that vaunted day when #4 comes into being, but I would rather see it gone for now.

andrew:
Agreed, should be pretty easy to add it back in later, for now 🔥

PR #235, opened 2017-01-11
Let users set blank refresh_interval to not have any auto refresh

### LINKED ISSUE #213: /settings won't let you set Notification Refresh Interval to 0
On instances with auto-refresh enabled, users can't set the interval to 0 disabling the feature.

The problem is [this min value](https://github.com/octobox/octobox/blob/25ac7d9fa5dff617caa22bb85568948683e55696/app/views/users/edit.html.erb#L8-L8). 

I will fix this when I have time later today.

WillAbides:
fixes #213 [/settings won't let you set Notification Refresh Interval to 0](https://github.com/octobox/octobox/issues/213)

This was a little bigger fix than I intended.  I ended up changing the value for no auto-refresh to be nil instead of 0.  That meant updating the schema.  There will still be users who have 0 values, so I overrode `user.refresh_interval` so that 0 maps to nil.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -30,7 +30,7 @@ class UsersController < ApplicationController
   end
 
   def update_user_params
-    if params[:user][:refresh_interval_minutes].present?
+    if params[:user].has_key? :refresh_interval_minutes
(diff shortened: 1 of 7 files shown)


andrew:
Thanks!

PR #240, opened 2017-01-13
Mark As Read on notification click

### LINKED ISSUE #239: Mark as Read on click, not sync
[Enhancement]

Currently, when you select an Unread issue/PR from the list and the new tab is opened, when you return to Octobox you will find that the issue/PR is still marked as Unread until the next sync action occurs.

It would be neat if we could sync the individual notification after opening (or mark as read from our end) on click to instantly reflect that the notification has indeed been read.

chrisarcand:
Closes #239

When a notification is clicked, instead of waiting to mark the notification as read on the next sync action, immediately mark the notification as having been read.

I dislike that this feature utilizes GET instead of PATCH for updating a notification but I know little to nothing about JQuery and see that this is how the toggling of 'starred' works, so... ¯\\\_(ツ)_/¯

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -129,6 +129,11 @@ function mute() {
   });
 }
 
+function markAsRead(id) {
+  $.get( "/notifications/"+id+"/mark_as_read");
+  $('#notification-'+id).removeClass('active');
+}
+
 function toggleArchive() {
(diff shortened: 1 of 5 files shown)


andrew:
Thanks! 👍

andrew:
Just deployed that 🚢

PR #245, opened 2017-01-15
Limit notification sync to 500 notifications

### LINKED ISSUE #242: Too many notifications synced on initial signup
Another side effect of https://github.com/octobox/octobox/pull/227 is that we pull in **all** the users notifications the first time they sign up, for very active github users this either takes ages, timeouts or crashes the rails app completely, need to add a limit back in!

WillAbides:
Fixes #242

This limits all notification downloads to 500 notifications (or whatever is set in `MAX_NOTIFICATIONS_TO_SYNC`)

Octokit doesn't allow you to limit the number of results you get, so I ended up[ patching its `paginate` method ](https://github.com/WillAbides/octobox/blob/d0f314c28e50d6fbe9cf511b7064c8dcc4df08f9/app/services/download_service.rb#L93-L122).  If this works out, I'll send an upstream PR with the change so we can get that code out of octobox.

I also extracted the download logic out of `Notification` into `DownloadService`.


andrew:
Lovely work!

sharwell:
:question: Do you have a recommended way to bring in all my unread issues instead of just the very small subset which this pull request limits it to?

chrisarcand:
@sharwell Officially, no. If you want to attempt to pull in all of them, I imagine you could set your user's `last_synced_at` to a very early date and use this API in a console (`DownloadService.new(user).fetch_unread_notifications`). This may or may not be so successful given an enormous number of notifications.

sharwell:
@chrisarcand Interesting. I'll have to put some thought into what my expectations might be around this and how to address them. I would not be surprised if the trigger that gets many users to try octobox is having way wayyyyyyyyyyyyyyyyyyyyyyyyyyyy more than 500 notifications pending and addressing the needs of those users is valuable but of course challenging.

chrisarcand:
@sharwell For sure, I'm one of those users. I had many many many thousands of notifications when I first got going (Pretty sure I was the user that first brought up the pain in #242). My standing is just that yeah, past history of notifications can't be brought in very well but the coincidence of not having an old notification decreases over time as you use octobox. ¯\\\_(ツ)_/¯ Not ideal, for sure, so if you have any ideas for how to 'onboard' a little better, we're all ears! 😄

PR #249, opened 2017-01-17
Fix configuration for rubucop to be compliant with new version

michaelrauh:
I noticed that running rubocop was producing an exception related to a change in supported methods in the rubocop.yml. This updates that out so that it will run without exception. 

```shell
Error: The `Style/MethodCallParentheses` cop has been renamed to `Style/MethodCallWithoutArgsParentheses`.
(obsolete configuration found in /Users/michaelrauh/Dev/octobox/.rubocop.yml, please update it)
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -716,7 +716,7 @@ Style/LineEndConcatenation:
                  line end.
   Enabled: false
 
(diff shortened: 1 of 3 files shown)


andrew:
Thanks!

PR #255, opened 2017-01-18
Allow optional access restriction to Octobox

tarebyte:
If you are running a private Octobox instance for your company, you may want to limit it to a GitHub organization or team.

This PR allows you to turn on restricted access via environment variables so that if you don't want someone to have access to your Octobox instance they can't.

This requires the use of the `read:org` scope to check for org and team membership.

Let me know what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,17 @@ PERSONAL_ACCESS_TOKENS_ENABLED=1
 
 Once that is set, users can set a personal access token on the Settings page (found on the user drop-down menu).
 
+## Limiting Access
+You can restrict access to your Octobox instance, and only allow members or a GitHub organization or team.  To limit access set the environment variable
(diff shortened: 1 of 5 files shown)


andrew:
@tarebyte I can't see where you're setting the `read:org` scope, is it just for the personal access token or does it need to go into the omniauth scope as well? Would be good to have this documented in the readme too

tarebyte:
@andrew apologies, if you use this feature then you would have to update your `GITHUB_SCOPE` environment variable to be `notifications,read:org`

andrew:
So this isn't aimed at working in combination with the personal access token workflow?

WillAbides:
@andrew: That's correct.  This is so we can run octobox at our place of employment and have it limited to only users in our employer's org.

andrew:
@WillAbides cool, LGTM just need the readme docs and we can merge 👍

WillAbides:
I added to README

https://github.com/octobox/octobox/blob/facb5a69c8a2fd79cd8af7dad59ef3dfc59401a5/README.md#limiting-access

chrisarcand:
Nice work @tarebyte 👍

PR #256, opened 2017-01-18
Enable web concurrency for Octobox in production

tarebyte:
This enables concurrency in the application for production, and allows us to kill puma workers on a rolling restart if they start taking up too much memory.

Thoughts?

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -21,7 +21,7 @@ environment ENV.fetch("RAILS_ENV") { "development" }
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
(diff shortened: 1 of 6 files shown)


config/puma.rb
```diff
@@ -21,7 +21,7 @@
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
+workers ENV.fetch("WEB_CONCURRENCY") { 1 }
```
tarebyte: I set this to be 1 by default so that we don't have to set anything special in dev.

andrew:
LGTM 👍

andrew:
Just deployed this to heroku, with `WEB_CONCURRENCY=2`, everything seems to be working well 🚂

PR #257, opened 2017-01-18
Add rake task for viewing application stats

tarebyte:
Worked with @WillAbides to make a neat little Rake task that gives you the some stats on your application.

`rake octobox:stats`

```
+------------------------------------------+
| Users: 1                                 |
| Users active within the last 24 hours: 0 |
| Notifications: 562                       |
+------------------------------------------+
```

Thoughts?

### DIFF
--- a/lib/tasks/stats.rake
+++ b/lib/tasks/stats.rake
@@ -1,5 +1,5 @@
 namespace :octobox do
-  desc "Get overall stats for application"
+  desc "Get overall stats for the application"
   task stats: :environment do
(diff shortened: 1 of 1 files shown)


andrew:
Nice, perhaps it can say the period that is considered recently active, i.e. Users active with the last 24 hours

tarebyte:
> Nice, perhaps it can say the period that is considered recently active, i.e. Users active with the last 24 hours

Good call, updated the text in 29c3f5b

PR #260, opened 2017-01-19
Fix type in environment variable interpolation for GITHUB_ env

tarebyte:
For some reason the tests didn't pick up on this. Will need to investigate.

**Update:** turns out no matter if we failed our not we redirect to root. I've added tests to make sure there isn't a `flash[:error]` on redirect.


chrisarcand:
Whoops, my bad. Good catch.

PR #267, opened 2017-01-19
Favicon unread count

### LINKED ISSUE #211: Update Favicon on Refresh with Unread Count/Indicator
I have 6 tabs that are pinned within Chrome, and one of them is now Octobox, one of the issues that I have is I need to keep checking the tab to see if there are any new notifications etc. 

It'd be great if the tab updated with a count of the total unread notifications, or an indicator at least. There should be two options to users, to remove this indicator either when the tab is viewed, or for the indicator to always be persistent based on the notification status. 

I sometimes like to leave them unread, but might not want the status indicator to be there all the time. 

If no one gets round to it, I'd potentially look at implementing something, but my rails knowledge is limited!

Thanks. <3

karpiu:
Fixes #211 

Feedback appreciated.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -31,6 +31,47 @@ function getMarkedOrCurrentRows() {
   return hasMarkedRows() ? getMarkedRows() : getCurrentRow()
 }
 
+function updateFavicon() {
+  console.log("update")
+  $.get( "/notifications/unread_count", function(data) {
+    var unread_count = data["count"];
+    if ( unread_count > 0 ) {
+      var old_link = document.getElementById('favicon-count');
+      if ( old_link ) {
+        $(old_link).remove();
+      }
+      
+      var canvas = document.createElement('canvas'),
+        ctx,
+        img = document.createElement('img'),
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/application.js
```diff
@@ -31,6 +31,47 @@ function getMarkedOrCurrentRows() {
   return hasMarkedRows() ? getMarkedRows() : getCurrentRow()
 }
 
+function updateFavicon() {
+  console.log("update")
```
tarebyte: Leftover from development 😄

karpiu: Nice find :+1:

andrew:
Since we're using turbolinks and effectively refreshing the page every time, could we just embed the unread count in the page and read it from there instead of making an Ajax call?

karpiu:
@andrew some actions does not reload page, like "mark as unread". We would still like to show the change in the favicon after this action immediately (at least I would be expecting such behaviour).

andrew:
Ok cool, LGTM then 👍

tarebyte:
Thanks @karpiu!

PR #272, opened 2017-01-19
Remove Vagrant reference

chrisarcand:
Now, I'm not really a Vagrant user so could be missing something entirely, but I don't see any actual Vagrant support in the project.

I'm sure it will be added by someone eventually, but we should probably remove the mention of the included Vagrant setup for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ For other updates, follow me on Twitter: [@teabass](https://twitter.com/teabass)
 
 ### Getting Started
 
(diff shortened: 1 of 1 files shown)


tarebyte:
We do have docker, so maybe we can update the text to point to the docker section?

chrisarcand:
@tarebyte I've got a larger PR reorganizing the docs coming, I'll take care of it there 💙

chrisarcand:
Alright alright, I'll stop being lazy. Fixed!

PR #273, opened 2017-01-20
Made error page links open in new tabs

GiacomoLaw:
The link to the issue page opened in the same tab, so I made it open in a new tab so the user can go back to Octobox on the same page.


andrew:
Thanks!

Issue #270, opened 2017-01-19
Auto sync


GiacomoLaw:
Instead of manually updating it, auto synchronising every so often would be great.

GiacomoLaw:
Just seen this:

https://github.com/octobox/octobox/blob/34c487d160c0826f6457d11fbe2479ac3758cce6/README.md#sync-notifications-automatically

Does it do this on octobox.io?

WillAbides:
I think that #200 is actually closer to what you want because it will update a loaded page instead of just running in the background.  Unfortunately neither of these are running on octobox.io.

Looking back at #200 I realize I neglected to update README.

GiacomoLaw:
@WillAbides I'll add that it doesn't auto sync on the hosted version.

GiacomoLaw:
Here is the README.md update: https://github.com/octobox/octobox/pull/271

GiacomoLaw:
Here are two potential fixes:

- https://github.com/octobox/octobox/pull/275
- https://github.com/octobox/octobox/pull/274

WillAbides:
I added doc for #200 here:  https://github.com/octobox/octobox/blob/master/README.md#allowing-periodic-notification-refreshes

I don't think that #274 or #275 add anything to isn't already there, but they do force a periodic page refresh on all users instead of making it optional.  I know that @andrew is opposed to forcing refreshes on users.

GiacomoLaw:
@WillAbides I think an option should be added, if someone could add that it would really be appreciated. I'm not good enough at ruby to do that. Sorry!

WillAbides:
@GiacomoLaw: It's not a matter of adding any new code.  Unless I'm missing something significant, I think Octobox already does what you want if it is enabled on the server.

So you can either self-host octobox or lobby @andrew to set `MINIMUM_REFRESH_INTERVAL` on octobox.io.

GiacomoLaw:
Okay, cool. Surely and option could be added?

chrisarcand:
> Okay, cool. Surely and option could be added?

@GiacomoLaw IIRC These options are disabled due to a current desire to *not* store user authentication details (understandable). Some work should be done before this option is enabled, at least (I think it's currently stored in plaintext). 

I would suggest self-hosting for now. A 'hobby' tier (free) on Heroku is enough to meet your needs and is easy to set up (I use this, and haven't had any need to upgrade it yet). 

> About self hosting, I find that it's public on Heroku, and I can't update from this repo.

[...]

PR #274, opened 2017-01-20
Potential fix for #270 - auto refresh

GiacomoLaw:
I've simply added:

`<meta http-equiv="refresh" content="120" />`

This automatically refreshes the page every 120. I know this works in html, so I dumped it in a head tag, but I'm not familiar with `.html.erb`, so I don't know if it works.

Hope this helps! It's just a temp fix, but shouldn't be too hard. 😄

### DIFF
--- a/app/views/notifications/index.html.erb
+++ b/app/views/notifications/index.html.erb
@@ -1,3 +1,6 @@
+<head>
+  <meta http-equiv="refresh" content="120" />
+</head>
 <div class="flex-content">
   <div class="flex-sidebar">
     <div class="flex-container">


GiacomoLaw:
I'm pushing it to Heroku to give a demo.

GiacomoLaw:
How can I request reviews? I don't see a button for some reason 😕

GiacomoLaw:
Here is another potential fix: https://github.com/octobox/octobox/pull/275

GiacomoLaw:
Already on there, just not enabled.

> It's not a matter of adding any new code. Unless I'm missing something significant, I think Octobox already does what you want if it is enabled on the server.
> 
> So you can either self-host octobox or lobby @andrew to set MINIMUM_REFRESH_INTERVAL on octobox.io.

PR #275, opened 2017-01-20
fix for #270 - added another way to refresh page

GiacomoLaw:


GiacomoLaw:
Already on there, just not enabled.

> It's not a matter of adding any new code. Unless I'm missing something significant, I think Octobox already does what you want if it is enabled on the server.
> 
> So you can either self-host octobox or lobby @andrew to set MINIMUM_REFRESH_INTERVAL on octobox.io.

PR #281, opened 2017-01-20
Clicking on an active filter should remove it

andrew:
I find myself clicking on the active filter to try and toggle it, this pr fills in that behaviour so my clicks won't be in vain.

screen shot 2017-01-20 at 5 43 14 pm

### DIFF
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -100,7 +100,7 @@ module NotificationsHelper
         block.call
         if active
           concat content_tag(:span, octicon('x', :height => 16), class: 'label text-muted')
-        else
+        elsif count.present?
           concat content_tag(:span, count, class: 'label label-muted')
         end
       end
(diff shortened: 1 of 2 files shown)


app/helpers/notifications_helper.rb
```diff
@@ -91,4 +91,19 @@ def filter_option(param, &block)
       end
     end
   end
+
+  def filter_link(param, value, count, &block)
```
WillAbides: 👍  moving this logic to the helper

WillAbides:
As an aside to this, I think it would make more sense for the filter count displayed to be the number of notifications you would see if you added that filter to your current filters.

andrew:
@WillAbides agreed, that needs some funky active record work at the controller level to get proper elasticsearch-style faceted filter counts

andrew:
screen shot 2017-01-20 at 7 06 27 pm

Removing the filters on repos and orgs works now, although there's some slightly horrible duplication in the helpers that needs further refactoring 😬

PR #284, opened 2017-01-21
License notice

### LINKED ISSUE #278: Clearer source link
From [the license](https://github.com/octobox/octobox/blob/f3101457f174c0ad56166f8841a2a5f40b21f045/LICENSE.txt#L650-L656):

>   If your software can interact with users remotely through a computer
network, you should also make sure that it provides a way for users to
get its source.  For example, if your program is a web application, its
interface could display a "Source" link that leads users to an archive
of the code.  There are many ways you could offer source, and different
solutions will be better for different programs; see section 13 for the
specific requirements.

Octobox almost does this through the 'Octobox support' menu link to GitHub issues, but it could do better by adding a link to the code and labeling it 'Source'.

I would also like to make the link configurable through an environment variable so that forks can comply without having to modify that part of the source.

WillAbides:
closes #278

This adds the license notice "© 2017 Andrew Nesbitt, [source](https://github.com/octobox/octobox) available under [AGPL 3.0](https://github.com/octobox/octobox/blob/master/LICENSE.txt)"  to the bottom of each page.  It also adds a `SOURCE_REPO` configuration option that can be used to link to an alternate repo in case it is run with modified source code.

My first choice for the layout was to render the footer at the [bottom of application.html.erb](https://github.com/WillAbides/octobox/blob/fa42c985c68ecd787d0f5fe08ffb9cf9bbfed3ed/app/views/layouts/application.html.erb#L42), but I couldn't find a good way to do that without it stealing real estate from notifications.  That would get quite annoying on mobile.  Instead I rendered it at the bottom of `notifications/index.html.erb` as well as `pages/home.html.erb` and `users/edit.html.erb`.

Logged Out:
2017-01-21 at 12 43 pm

Notifications:
2017-01-21 at 12 40 pm

User Settings:
2017-01-21 at 12 57 pm

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -183,3 +183,8 @@ To get a team's id:
 `curl https://api.github.com/orgs/<org_name>/teams`.
 You must be authenticated with access to the org. This will show you a list of the org's teams. Find your team on the list and copy its id
 
+## Source Link for Modified Code
+
+If you have modified the Octobox code in any way, in order to comply with the AGPLv3 license, you must link to the modified source.  You
+can do this by setting the `SOURCE_REPO` environment variable to the url of a GitHub repo with the modified source.  For instance, if 
+you run this from a fork in the 'NotOctobox' org, you would set `SOURCE_REPO=https://github.com/NotOctobox/octobox`.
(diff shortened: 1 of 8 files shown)


andrew:
Wonderful, a model pull request too, thanks!

PR #286, opened 2017-01-22
Configuration overhaul

### LINKED ISSUE #280: Configuration API
I'd like to write a simple configurator to allow people to configure the application in a single convenient spot, with more flexibility, rather than the bucket of environment variables that we currently have.

I was thinking a configuration block available from the top namespace; we can then gitignore a specifically named initializer that people can add to configure their box how they want without tracked code changes.

A few questions up for discussion:

Concerning current environment variables...
* Can some of the current environment variables be immediately removed?
* If they can't be, should they be deprecated? (add a warning wrapper that will yet at you if you use them, say that they'll be removed 'eventually'?)
* Or, should we just keep the existing ones any new configuration settings will just have a configurator option with no env var option?

Lastly, if some variables *are* removed/deprecated, are there any current ones which, even if you think there was a configurator option, you'd still really like to see as an environment variable? I haven't used the personal access token feature, but perhaps that one is nice to just be able to do without writing any new file (unsure).

chrisarcand:
Closes #280 

This PR consolidates non-sensitive Octobox-specific configuration into a single place, accessible to read and write via block form from `Octobox.config`. Options that were only ENV vars before are now programmatically accessible as well. It has a number of advantages that I see:

* Configurable options are all found in the same spot (`lib/octobox/configurator`) All globally defined helpers are all found in the same spot (`lib/octobox`, for now)
* All application callers use the configuration object as the one source of truth. No individuals are calling out to environment variables on their own (all env vars are respected and handled in the configuration alone)
*  The namespace and configuration is loaded as the app is being initialized; thus, no awkward naming and ordering of initializers (ex: Octokit needing GitHub domain urls, etc)
* With an initializer file gitignored (`config/initializers/custom.rb`), you can make your own untracked initializer in your instance if you prefer to configure it there.
* Some other reasons that I thought of last night that I can't remember now :)

**Note:** This doesn't change behavior around environment variables, as discussed in #280. TBH I was the one wanting programmatic setting and even I actually think it's more appropriate to have methods defer to ENV (we currently allow overriding of ENV).

r? @WillAbides


lib/octobox.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def restricted_access_enabled?
+      config.restricted_access_enabled
+    end
+
+    def contributors
+      @contributors ||= Octokit::Client.new(auto_paginate: true).contributors(Octokit::Repository.from_url(config.source_repo))
+    rescue
+      nil
+    end
```
chrisarcand: @WillAbides Note this is an actual behavior change; I thought it weird to have contributors configurable so there's no writer here, it's just a helper. I can change it back if you had some thoughts on that.

lib/octobox/configurator.rb
```diff
@@ -1,23 +1,30 @@
...
+
+    def github_api_prefix
+      return @github_domain_api_prefix if defined?(@github_domain_api_prefix)
+
+      if github_domain != 'https://github.com'
+        @github_domain_api_prefix = "#{github_domain}/api/v3"
+      else
+        @github_domain_api_prefix = "https://api.github.com"
+      end
+    end
```
chrisarcand: Review these two methods carefully; they were not configurable beyond env before (due to initializer timing concerns) so they're slightly edited. If I understand correctly the api_prefix is entirely dependent on the domain so there's no writer here intentionally.

lib/octobox/configurator.rb
```diff
@@ -34,20 +41,21 @@ def restricted_access_enabled
...
+      id = @github_organization_id || ENV['GITHUB_ORGANIZATION_ID']
+      return id.to_i if id.present?
     end
+    attr_writer :github_organization_id
+
+    def github_team_id
+      id = @github_team_id || ENV['GITHUB_TEAM_ID']
+      return id.to_i if id.present?
+    end
+    attr_writer :github_team_id
```
chrisarcand: There are also slight changes in this area; the configuration is now returning an actual integer so the callers don't have to call `to_i` every time.

lib/octobox.rb
WillAbides: You're right.  That was unnecessary.

app/controllers/sessions_controller.rb
```diff
@@ -54,12 +54,12 @@ def team_member?(*args)
   end
 
   def member?(client, nickname, type:)
-    id = ENV["GITHUB_#{type.to_s.upcase}_ID"]
+    id = Octobox.config.public_send("github_#{type}_id")
```
WillAbides: I'm not a big fan of using public_send when a case or if statement will work, I would suggest changing it, but there's another a few lines down so we may as well be consistent.

lib/octobox.rb
```diff
@@ -0,0 +1,31 @@
...
+module Octobox
+  class << self
+    def config
+      @config ||= Configurator.new
+      if block_given?
+        yield @config
+      end
+      @config
+    end
+
```
WillAbides: How about adding method_missing so we don't need to use `Octobox.config`?

```ruby
    def method_missing(method, *args)
      config.public_send(method, *args) if config.respond_to?(method)
    end
```

config/initializers/octobox.rb.example
```diff
@@ -0,0 +1,58 @@
+###
+# Octobox can be configured using environment variables or via it's
+# configuration API directly. To use this file, copy it to config/initializers/octobox.rb
+# Changes in your new configuration file will not be tracked by Git.
+#
+# Note that all of these options have a corrresponding environment variable
```
chrisarcand: Corarghresponding yo ho ☠️

lib/octobox.rb
chrisarcand: I actually kind of like the separation from the top level module, was thinking ala [`Rails.configuration`](https://github.com/rails/rails/blob/d25fe31c40928712b5e08fe0afb567c3bc88eddf/railties/lib/rails.rb#L42-L45). I'm not terribly passionate on it though, so if you prefer delegating it, 👌

chrisarcand: @andrew Got a preference?

chrisarcand:
> Instead of repeating the descriptions of each config option in INSTALLATION.md and octobox.rb.example

Yeah, was starting to think that myself as I was finishing it 😕  Good call, I'll do that.

Will finish this up tomorrow morning. 👍

lib/octobox.rb
WillAbides: I don't have a good reason to want it.  It probably is better to have it separate

chrisarcand:
@WillAbides Ready for another look; I decided to leave the delegation out for now, at least. Just doc changes.

INSTALLATION.md
```diff
@@ -106,6 +109,30 @@ Octobox will be running on [http://localhost:3000](http://localhost:3000).
...
+
+```ruby
+Octobox.config do |c|
+  # Example: Enabling personal access tokens
+  c.personal_access_tokens_enabled = true
+end
+```
+
+A sample configuration block with all available options can be found in
+`config/initializers/octobox.rb.sample`. Note that each environment variable\*
```
WillAbides: You mean octobox.rb.example

INSTALLATION.md
```diff
@@ -106,6 +109,30 @@ Octobox will be running on [http://localhost:3000](http://localhost:3000).
...
+mentioned in this guide has a corresponding option in the configuration block.
+
+To use and customize these options, copy `octobox.rb.sample` to `octobox.rb`.
```
WillAbides: octobox.rb.example here too

INSTALLATION.md
chrisarcand: Doh thanks

chrisarcand:
Fixed

WillAbides:
Since this is a significant change, maybe we should see what @andrew or @tarebyte have to say about it before merging.

andrew:
Feels weird to have this pattern, which I've seen in lots of ruby libraries, in an application, probably because if you need to do arbitrary ruby code in your application's config then something's missing from the app itself.

Was going to mention that the whole config block feels like a code smell but it's already a grab bag of personal tweaks and hacks toggled by env vars, as long as it stays out of the happy path then I'm not too fussed about it.

Other than that, as long as it doesn't break existing installs when they fetch master then it's all good 👍

tarebyte:
I'm not a huge fan of having an example file and then asking users to check in the real file.

I'd rather just point and say, if you want the options uncomment this ➡️

chrisarcand:
@andrew I see your point, and while I think I'd still do this I think that's a good argument for reversing the priority of these options to defer to environment variables (in the current config file the env vars can be overruled)

chrisarcand:
@tarebyte I'm not asking the users to check in the initializer they make at all - in fact, the whole point is to avoid any tracked changes so checking it in would defeat the purpose.

chrisarcand:
I'm also completely open to not making such a fuss about it in the README and not having an example file, just leaving /octobox.rb ignored as your own initializer you can do anything you want with. That would alleviate the awkwardness from both comments above, perhaps. I fully understand that 90% of users will probably just use env vars :)

chrisarcand:
After more consideration and some excellent feedback, I've decided to make this less awkward feeling with all the same benefits:

* Leaving the configuration bit more open ended in the documentation. Instead of a 'configuration feature', just a mention that you can add a custom initializer to do whatever you want with outside of the annoyance of Git tracking, using the accessible Octobox configuration as an example.
* Ignored initializer name changed to `custom.rb` to avoid any desire to want to actually use `octobox.rb` for our own purposes later.
* Removed initializer template. I, too, would have preferred @tarebyte's desire to have a configuration that you just uncomment as you want but the whole point here is to customize without tracking. And if you're really looking to configure through this API you probably already know what you're looking for/doing anyway. 

I think this direction reaps all the benefits in the original description without making it seem like some awkward feature that we have to maintain, which is sort of the vibe this was emitting originally.

andrew:
👏

PR #287, opened 2017-01-23
WIP Javascript Tests

WillAbides:
This is a work in progress.  I am posting to get some early feedback.

This allows us to test Octobox's javascript. It uses [Mocha](https://github.com/mochajs/mocha) as the test framework and [Teaspoon](https://github.com/jejacks0n/teaspoon) to run the tests. [Magic Lamp](https://github.com/crismali/magic_lamp) lets us test against our Rails templates with test data.

So far, the one and only actual test is in [`application_spec.js`](https://github.com/WillAbides/octobox/blob/3b5360142b1329545d57b7bf00bf045cfc309e57/spec/javascripts/application_spec.js). It's just a placeholder to have something to run until some useful tests are written.

Tests can be run locally with `bundle exec teaspoon`.  Phantomjs will be set up on the first run. 

TODO:
- [ ] DRY up test data so it isn't duplicated in `/test` and `/spec`
- [ ] See if we can get rid of `/spec` altogether and run everything from `/test`
- [ ] Document how to write and run javascript tests
- [ ] Write some useful tests
- [ ] Get MagicLamp out of routes.rb


chrisarcand:
I'm far from an authority on JS testing but lib choices and TODO look good to me. 

Only comment so far is adding to TODO: Let's add a javascript test task and separate the command from the implementation. Expect `rake test:javascript` to execute JS tests

config/routes.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 Rails.application.routes.draw do
+  mount MagicLamp::Genie, at: '/magic_lamp' if defined?(MagicLamp)
```
andrew: I hate that the test framework is leaking into the application logic itself here, is there any way we can inject this from the test env instead?

spec/javascripts/application_spec.js
```diff
@@ -0,0 +1,10 @@
+
+describe('header', function() {
+  beforeEach(function() {
+    MagicLamp.load("layouts/header");
+  });
+
+  it ('has the right avatar src', function() {
+    expect($("a.dropdown-toggle img").attr("src")).to.equal("https://github.com/foo.png?size=40");
```
andrew: Isn't this weird to test here as the avatar is set in html by the rails app?

spec/javascripts/fixtures/users.yml
```diff
@@ -0,0 +1,29 @@
+# Read about fixtures at http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
+
+andrew:
+  github_id: 1060
+  access_token: <%= SecureRandom.hex(20) %>
+  github_login: 'andrew'
+
+tokenuser:
```
andrew: Feels like we'd get a lot of mileage switching to factory girl now with all these very similar fixtures

spec/javascripts/application_spec.js
WillAbides: It's definitely not going to make it out of WIP. This is the first thing I saw on the page when I was trying out how Magic Lamp works, and I just kept it as a placeholder until I write some real tests.

config/routes.rb
WillAbides: I hope there's a better way.  This is copied from Magic Lamp's README.  I'm getting it to work by the book first before making it nice.  I'll add this to TODO so I don't forget and leave it here.

WillAbides:
I don't have time to finish this right now.  I'm closing it to that it doesn't sit here like a Camaro on blocks in the front yard.

If anybody wants to pick up where I left off, I'll keep the branch around at https://github.com/WillAbides/octobox/tree/js-tests

chrisarcand:
> I'm closing it to that it doesn't sit here like a Camaro on blocks in the front yard.

💯 Beautiful analogy. 👏😆

PR #292, opened 2017-01-24
User factory girl for test fixtures

WillAbides:
This gets rid of the yml fixtures and replaces them with factory girl fixtures


app/controllers/users_controller.rb
```diff
@@ -5,7 +5,9 @@ class UsersController < ApplicationController
   def edit; end
 
   def update
-    if current_user.update_attributes(update_user_params)
+    params = update_user_params
+    user = current_user
```
andrew: that assignment seems unnecessary

andrew:
👍

app/controllers/users_controller.rb
WillAbides: Thanks.  It's totally unnecessary.  I split it out to debug something and accidentally left it in.

WillAbides:
I messed this up by having a local .env file with `PERSONAL_ACCESS_TOKENS_ENABLED=true` and `MINIMUM_REFRESH_INTERVAL=2` so a lot of tests are failing on ci that don't fail locally.  I'll push a fix later today.

WillAbides:
It's fixed and ready to go now

andrew:
👏

PR #298, opened 2017-01-26
Made source and licence open in tab

GiacomoLaw:
```
source_link = "<a href='#{Octobox.config.source_repo}' target='_blank'>Source</a>"
    license_link = "<a href='#{Octobox.config.source_repo}/blob/master/LICENSE.txt' target='_blank'>AGPL 3.0</a>"
```

I added`target='_blank'` to the source and license link. I believe this is the way to do it, but note the standard way is `target="_blank"`. I did it with `'` as that is how the links are added, and using `"` would cause a conflict.

Thanks, hope this helps!

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -31,8 +31,8 @@ module ApplicationHelper
   end
 
   def license_message
-    source_link = "<a href='#{Octobox.config.source_repo}'>Source</a>"
-    license_link = "<a href='#{Octobox.config.source_repo}/blob/master/LICENSE.txt'>AGPL 3.0</a>"
+    source_link = "<a href='#{Octobox.config.source_repo}' target='_blank'>Source</a>"
+    license_link = "<a href='#{Octobox.config.source_repo}/blob/master/LICENSE.txt' target='_blank'>AGPL 3.0</a>"
     "#{source_link} available under #{license_link}".html_safe
   end
 
--- a/test/helpers/application_helper_test.rb
+++ b/test/helpers/application_helper_test.rb
@@ -21,13 +21,13 @@ class ApplicationHelperTest < ActionView::TestCase
 
   test 'license_message has correct link when SOURCE_REPO is not set' do
     ENV.stubs(:[]).with('SOURCE_REPO').returns(nil)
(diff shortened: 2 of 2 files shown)


andrew:
👍  you'll need to update these two tests as well: https://github.com/octobox/octobox/blob/3f9fe625b44c1d072a8eefb67438cbf645aa1c6e/test/helpers/application_helper_test.rb#L22-L30

GiacomoLaw:
@andrew I added it - is that correct?

andrew:
@GiacomoLaw yeah just need to add target blank to the "Source" links in the tests as well on the same lines

GiacomoLaw:
Okay, sorry.

GiacomoLaw:
There you go, hope that works. 😬

andrew:
Yep, tests pass, thanks very much!

WillAbides:
@GiacomoLaw: I think you mentioned somewhere that you are new to Ruby.  Do you know how to run tests locally using rake?  If not, I can give you some pointers.

GiacomoLaw:
@WillAbides Often I just host it on Heroku and test it from there. Pointers would be awesome, thank you! 😄

WillAbides:
I created #300 to put this in documentation, but in the mean time here are the basics.

You need to have two things installed ahead of time:  Ruby version 2.4.0 and PostgreSQL. This part is pretty OS dependent, and I don't know what operating system you are using.  If you need help getting either installed, let me know what issues you are having. There is a good chance I can help you if you are using Mac or Linux. My Windows knowledge is pretty rusty though.

Once you have those installed and postgres is running:

- Verify ruby is correctly set up in your path by running `ruby -v` in the octobox directory.  The output should start with `ruby 2.4.0p0`
- Install the bundler gem by running `gem install bundler`
- Run `bundle install` to install all the gems in Gemfile
- Set up your database by running `bundle exec rake db:create db:migrate`

Now you should be able to either run locally or run tests.

- To run all tests: `bundle exec rake test`

- To run a local instance `bundle exec rails server`

GiacomoLaw:
@WillAbides Thanks so much! I use Windows, I'll give it a go. Thank you!

PR #301, opened 2017-01-26
fixed #291 - add fixed white space between footer elements

### LINKED ISSUE #291: Footer license/copyright looks weird on big screens
screen shot 2017-01-24 at 9 33 29 am

Only just saw the updated footer on desktop, and the stretchy spacing looks really off on large screens.

Mostly leaving this issue to bug myself to fix it 😬

GiacomoLaw:
Here is a demo: http://codepen.io/GiacomoLaw/pen/OWQYBz

Hope this helps!

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -73,7 +73,7 @@ html {
   display:inline-block;
 }
 
-.container {
+.footer-spread {
   align-items: center;
   display: flex;
   justify-content: center;
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -80,7 +80,7 @@ html {
 }
 
 #left {
-  margin-right: 80px;
+  margin-right: 40px;
 }
 
 #centre {
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -79,11 +79,11 @@ html {
   justify-content: center;
 }
 
-.left {
+#left {
   margin-right: 40px;
 }
 
-.centre {
+#centre {
   margin-right: 40px;
 }
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -73,6 +73,20 @@ html {
   display:inline-block;
 }
 
+.container {
+  align-items: center;
+  display: flex;
+  justify-content: center;
+}
+
+#left {
+  margin-right: 80px;
+}
+
+#centre {
+  margin-right: 40px;
+}
+
 @media (max-width: $screen-xs-max) {
   .footer-data {
     width: unset;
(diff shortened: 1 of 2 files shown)


GiacomoLaw:
Going to improve this.

WillAbides:
Instead of adding style to the html, I would recommend changing the [css here](https://github.com/octobox/octobox/blob/f04baafa5150218ff2c8f82014a4ca20249d63fb/app/assets/stylesheets/application.scss#L63-L84).  If the left and right sides need different formatting, you could split `.footer-data` into two classes `.footer-data-left` and `.footer-data-right`.

You may already know this, but you can use [chrome dev-tools](https://developers.google.com/web/tools/chrome-devtools/inspect-styles/) preview css changes on the live page.  That can dramatically speed up workflow on design work.

GiacomoLaw:
Just a quick question, does Ruby on Rails automatically refer to the stylesheet or do I have to add a link in the header myself?

GiacomoLaw:
I presume it does seeing as there are no links.

GiacomoLaw:
There you go, should work well.

andrew:
@GiacomoLaw can you uses classes for left and right instead of ids?

GiacomoLaw:
@andrew Sure, will do

GiacomoLaw:
@andrew Done. :)

GiacomoLaw:
I think it would be better if they were left as `ids`.

andrew:
@GiacomoLaw I just checked this out locally, looks like changing the `.container` class breaks the homepage:

screen shot 2017-02-01 at 11 11 08 am

GiacomoLaw:
Woops, let me just do something.

GiacomoLaw:
Okay, I reverted it to id.

It should be displaying like this: http://codepen.io/GiacomoLaw/pen/OWQYBz

GiacomoLaw:
Is that good? 😄

chrisarcand:
@GiacomoLaw I hate to be a Git history purist (and usually don't mind some extra commits), but would you mind squashing these commits into one before this is merged? 8 commits in history before I get back to the one that specifies why this small change was done is just a little too much IMO :)

If you need help doing that, [Steve Klabnik has an excellent short reference](http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request) showing how.

andrew:
The .container class isn't specific enough and still breaks the homepage, needs to be scoped to just the footer

GiacomoLaw:
Okay, will do. Is the .container class used elsewhere? I could just change the name to something else.

GiacomoLaw:
I renamed the container class to a more specific name. Should that work now?

Thanks! 🙂

andrew:
I changed the footer quite a bit in https://github.com/octobox/octobox/pull/346, going to close this one as I don't think it's relevant anymore

PR #304, opened 2017-01-29
Fix error when SOURCE_REPO is present but empty

WillAbides:
Having `SOURCE_REPO=` in `.env` was causing a 500 error.  This fixes that.


lib/octobox/configurator.rb
```diff
@@ -54,7 +54,8 @@ def github_team_id
     attr_writer :github_team_id
 
     def source_repo
-      @source_repo || ENV['SOURCE_REPO'] || 'https://github.com/octobox/octobox'
+      env_value = ENV['SOURCE_REPO'].blank? ? nil : ENV['SOURCE_REPO']
```
chrisarcand: Really interestingly, while I was writing tests for the other methods here I discovered that `ENV.fetch` seems to strip whitespace (`github_domain` above, using fetch, passes your test, too):

```ruby
> ENV['ASDF'] = ""
=> ""
> ENV['ASDF']
=> ""
> ENV.fetch('asdf', 'blah')
=> "blah"
> ENV['ASDF'] = "     "
=> "     "
> ENV.fetch('asdf', 'blah')
=> "blah"
```

Didn't know this, so I'll probably use that instead in my followup.

chrisarcand: 🤐 Forget that.

Yay for capitalization!

PR #308, opened 2017-02-05
Add support for updating beyond page size

HParker:
screen shot 2017-02-04 at 6 36 12 pm

Before this change you could select all visible items and update them
together, but if you wanted to update more than your current page
size, you had to select each page one at a time until you had finished
removing items from your current query.

Now you have the option to select beyond the set of currently visible
items on the page.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -15,7 +15,8 @@ function getMarkedRows(unmarked) {
 }
 
 function getIdsFromRows(rows) {
-  return $.map(rows, function(row) {return $(row).find("input").val()})
+  return $('button.select_all').hasClass('all_selected') ?
+    'all' : $.map(rows, function(row) {return $(row).find("input").val()})
 }
 
 // returns true if there are any marked rows (or unmarked rows if unmarked is true)
@@ -39,7 +40,7 @@ function updateFavicon() {
       if ( old_link ) {
         $(old_link).remove();
       }
-      
+
       var canvas = document.createElement('canvas'),
         ctx,
         img = document.createElement('img'),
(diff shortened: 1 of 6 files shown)


WillAbides:
This is a great feature, and I really like the changing of the emphasized number to indicate what is selected.  I haven't looked at the code yet, but I do have some usage notes.

- I wonder if we could merge this with select-all to make it a three way toggle between these three:
2017-02-06 at 8 45 am
2017-02-06 at 8 57 am
2017-02-06 at 8 58 am

- If the buttons aren't merged, the new button should be moved to be directly next to the primary select-all checkbox to keep the selecting functionality together.

- The clipboard icon is a bit confusing, a user might think the button copy notifications instead of selecting.  Maybe ☑  would be better there.

- A tooltip would also be useful.  Maybe "Select all 28 notifications in scope" and "Select only the 20 displayed notifications"

chrisarcand:
For more inspiration, this sort of thing reminds me of Gmail's solution to this:

HParker:
Yeah, what do you all think? should we go with the gmail style, or the select 2 style?

Minimally we can just change to the ☑ and move it to the front of the list.

WillAbides:
> Minimally we can just change to the ☑ and move it to the front of the list.

I'd go with that and call it done

HParker:
screen shot 2017-02-10 at 2 18 58 pm

Alright I ended up going with the octicon and moved it to the front.
I hope this looks like you would expect.

HParker:
Cool, I added a tooltip. If you like it this way, I can squash these down to 1 commit.

andrew:
Thanks @HParker, I squashed and merged it from the PR interface, going to ship it to octobox.io now 🚂

HParker:
Thanks Andrew! Have a wonderful Monday!

chrisarcand:
@HParker Circling back around to say that I've been using this a lot and am extremely happy with the UI experience you decided on, thanks!

HParker:
Thanks @chrisarcand that made my day.

PR #313, opened 2017-02-18
Click on row to select it

technicalpickles:
I find myself often doing this:

- visually skimming a list of notifications
- click an issue to browse
- come back to octobox and use x to select the row, and e to archive

When I do that though, the row I just clicked on isn't the current one, so I end up reading some unrelated (and possible unread) notification.

By adding a click handler for notification rows, this means that
clicking anywhere in the row, be it on a td or a link to an issue,
starring, the clicked row will be set as the new current.

Also worth mentioning this is consistent with how navigation works in gmail, when you click into something and come back, the message you just viewed is made the current.

I did not see any tests for javascript stuff, or I would have added one.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -32,6 +32,14 @@ function getMarkedOrCurrentRows() {
   return hasMarkedRows() ? getMarkedRows() : getCurrentRow()
 }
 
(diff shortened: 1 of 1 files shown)


andrew:
Thanks, I'll definitely find this useful 👍

PR #314, opened 2017-02-21
Use new endpoints to determine team and org membership so that private memberships work

jules2689:
Problem
---
We have a lot of private members in our Org. ~~The used endpoint (https://developer.github.com/v3/orgs/#list-user-organizations) does not include private memberships.~~ This does not include the right scopes to read those.

Fix
---
~~Use the access token for the individual user to determine the membership instead. This lets us use https://developer.github.com/v3/orgs/#list-your-organizations and https://developer.github.com/v3/orgs/teams/#list-user-teams which will include private memberships.~~
- Add the `read:org` scope.
- Document the scopes
- Remove some dynamic programming that wasn't needed and confused me

Side Effects
---
The token must have the `read:org` scope, but only if `restricted_access` is enabled. So only add it if restricted access is being used. I've updated the docs to mention this where I saw fit and I added a section in INSTALLATION.md to talk about scopes as well.

cc @tarebyte


test/support/stub_helper.rb
```diff
@@ -11,26 +11,42 @@ def stub_env(env = {})
...
+  def stub_organization_membership_request(organization_id:, successful:)
+    orgs_url = %r{https://api.github.com/user/orgs}
 
     headers          = { 'Content-Type' => 'application/json', 'Cache-Control' => 'no-cach, no-store' }
-    default_response = { body: nil, headers: headers }
-
-    response = successful ? default_response.merge(status: 204) : default_response.merge(status: 404)
+    default_response = { headers: headers, status: 200 }
+
+    response = if successful
```
tarebyte: If the response wasn't successful, what do you think about returning an empty `Array` as the body?

app/controllers/sessions_controller.rb
```diff
@@ -37,34 +37,29 @@ def authorize_access!
     return true unless Octobox.restricted_access_enabled?
 
     client = Octokit::Client.new(access_token: auth_hash.credentials.token)
```
jules2689: Woah, how did you make the diff appear in a comment?

app/controllers/sessions_controller.rb
```diff
@@ -37,34 +37,29 @@ def authorize_access!
...
 
-  def member?(client, nickname, type:)
-    id = Octobox.config.public_send("github_#{type}_id")
-    return false unless id.present?
-
-    begin
-      client.public_send("#{type}_member?",
-        id,
-        nickname,
-        headers: { 'Cache-Control' => 'no-cache, no-store' }
```
jules2689: Added this to the new requests :+1:

test/support/stub_helper.rb
jules2689: Done.

app/controllers/sessions_controller.rb
tarebyte: So when you make a code block instead of using a language like `ruby` all you do is use `diff` instead 😄

test/support/stub_helper.rb
tarebyte: @jules2689 it looks like you might've missed the `stub_team_membership_requests` method but updated the `stub_organization_membership_request`

jules2689: how did I possibly do that? Also, side note. I forgot about this comment until octobox reminded me <3

app/controllers/sessions_controller.rb
```diff
@@ -36,35 +36,30 @@ def auth_hash
...
-      )
-    rescue Octokit::Error
-      false
-    end
+  def team_member?(client)
+    team_id = Octobox.config.github_team_id
+    return false unless team_id
+
+    teams = client.user_teams(headers: { 'Cache-Control' => 'no-cache, no-store' })
```
tarebyte: @jules2689 sorry had a thought after I approved this PR, what do you think about using this API instead?

https://developer.github.com/v3/orgs/teams/#get-team-membership

And then checking to make sure the state is `active`. This would save us from listing through all of the teams that user belongs to.

app/controllers/sessions_controller.rb
```diff
@@ -36,35 +36,30 @@ def auth_hash
...
   end
 
-  def organization_member?(*args)
-    member?(*args, type: :organization)
-  end
+  def organization_member?(client)
+    org_id = Octobox.config.github_organization_id
+    return false unless org_id
+
+    orgs = client.organizations(headers: { 'Cache-Control' => 'no-cache, no-store' })
```
tarebyte: So going over this again, we should revert to using the https://developer.github.com/v3/orgs/members/#check-membership here.

It is scoped to the user making the request so even if they are part of the org privately they should still be able check their own membership.

jules2689:
Ok, so the new org endpoint requires the _name_ of the org, not the _id_ of the org - the current requirement of this app. That means that either usage changes or we make another API call,

tarebyte:
> Ok, so the new org endpoint requires the _name_ of the org, not the _id_ of the org - the current requirement of this app. That means that either usage changes or we make another API call.

This was a bug in Octokit 😄 I've made a PR to fix that issue https://github.com/octokit/octokit.rb/issues/869

You could use the [`organization_member?`](https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/organizations.rb#L166-L190) method (API docs https://developer.github.com/v3/orgs/members/#check-membership) and that would save you some parsing.

Just a thought.

jules2689:
> You could use the organization_member? method (API docs https://developer.github.com/v3/orgs/members/#check-membership) and that would save you some parsing.

I'm not clear if that includes active only or active/pending, I assume it's active only?

tarebyte:
> I assume it's active only?

That is correct.

jules2689:
@tarebyte updates on https://github.com/octokit/octokit.rb/pull/869?

tarebyte:
> @tarebyte updates on octokit/octokit.rb#869?

I just merged it 😄 but I don't have access to cut a new release on ruby gems so for now you may just want to grab the commit hash.

jules2689:
@tarebyte  Ive pinned the ref to the latest. Once a new octokit is cut, we can put it back to the released version.

As for the suggestion in https://github.com/octobox/octobox/pull/314#issuecomment-281536303, this is a bit weird. This is the endpoint it used to use, not the one I originally listed (I have no idea how I got to that conclusion anymore). I'm not sure why this wasn't working for private members. I'm going to need to look at this again.

jules2689:
@tarebyte this is ready to go. I found out what the problem was in https://github.com/octobox/octobox/pull/314#issuecomment-282365753

Previously, those methods didnt work because the `read:org` scope was not added. Now that Octobox handles scopes properly, we can use the old endpoints again (and we come full circle).

i'll update the README with what this PR does now

Issue #39, opened 2016-12-18
API


andrew:
As mentioned in https://github.com/andrew/octobox/issues/17, we could build an API to make it easy to build different kinds of UI on top of the data layer.

andrew:
Closing in favour of https://github.com/octobox/octobox/issues/471, the one last piece to follow up on #320

PR #320, opened 2017-02-22
JSON API

jules2689:
What this does
---
This attempts to address https://github.com/octobox/octobox/issues/39

Taking a look at the current system, there are only a handful of endpoints. Notably:
 - `notifications#index`
 - the notifications actions
 - `user#profile`
 - `user#update`
 - `user#destroy`

The intention with the API is that the 2 get methods (`notifications#index` and `user#profile`) provide all information.
- All notifications actions will return a HEAD ok, so no changes are required there,
- `users#update` will provide either an OK, or a json blob with the errors and a 422 status.
- `users#destroy` just provides a HEAD OK 

Docs are provided as YARDoc style documentation above the public API methods. I have to check to use the proper keyword still.

Lastly, this *does not* handle any authentication mechanism dedicated to JSON. I suspect we could add an API token mechanism to the user profile and auto-generate a key on create. The key would be shown in the user's profile. This will be handled later.

cc @andrew @tarebyte


jules2689:
Any thoughts @tarebyte @andrew ?

andrew:
@jules2689 sorry totally missed this PR, looks great, we'll need an authentication method in before this is usable right?

andrew:
Just fixed the conflicts with the Gemfile.lock for you

jules2689:
@andrew yea, we need to get authentication of some sort in before it's useable as an actual API. I figured this could be done separately.

andrew:
@jules2689 ok cool, will merge this 👍

PR #322, opened 2017-02-23
Rename status to unread

jules2689:
Changes things to reflect the state in the DB and make it more easily understood by endusers:
`https://octobox.shopify.io/?status=true` becomes `https://octobox.shopify.io/?unread=true`


andrew:
Just deployed it to https://octobox.io 🚢

Issue #83, opened 2016-12-20
More information regarding author


WilsonChiang:
It would be nice if there was the option of showing the author of a PR/issue, and/or showing the person who last caused a notification to occur. If clutter is a concern maybe this could be a toggleable option.

If there's any interest for this I could work on adding this.

andrew:
Yeah would be good, related to https://github.com/andrew/octobox/issues/4

andrew:
First part of this has been merged in #411 and is available behind the `FETCH_SUBJECT` env var feature flag, docs here: https://github.com/octobox/octobox/blob/master/INSTALLATION.md#downloading-subjects

Still working on getting it available on https://octobox.io

andrew:
Closing this as it's available if you run your own instance, waiting on https://github.com/octobox/octobox/issues/472 for it to be enabled on https://octobox.io

PR #326, opened 2017-03-06
[POC] Display subject author

chrisarcand:
Potential solution for #83 that doesn't involve having to fully implement #4. 

*Note: Opened for early feedback. Don't merge this until I actually fix the tests and stuff for it.*

I've continued to wish we had authorship of the things I'm browsing through. Thus, here's a thing. When notifications are synced, the author of the subject of the notification is recorded by making an additional call to the subject (issue, commit, whatever) using the notification response. This is only done once per notification thread, as the author of the subject should never change and making an extra API per notification on sync would be terrible. 

I'm really pleased with the results:

Thoughts @andrew @WillAbides @tarebyte ? Thought I'd open this early for feedback before I write tests and whatnot. This implementation gives the oft requested authorship feature without having to fully implement and sync the concept of a subject (issue/commit, etc). I've been running this change on my own box today and really like it, will probably keep it on my fork regardless :)


app/models/notification.rb
```diff
@@ -67,6 +67,10 @@ def repo_url
     "#{Octobox.config.github_domain}/#{repository_full_name}"
   end
 
+  def subject_author_url
+    subject_author ? "#{Octobox.config.github_domain}/#{subject_author}" : ""
```
chrisarcand: 🙇  @davidcelis

app/assets/stylesheets/application.scss
```diff
@@ -413,6 +413,11 @@ td.keys {
         text-overflow: ellipsis;
       }
     }
+    .notification-subject-author:before {
+      content: "by ";
+      color: $text-muted;
+      font-size: 10px;
+    }
```
chrisarcand: I can't CSS, so like...feel free to throw tomatoes. 🍅  I just think the stacked nature of the repo/author on mobile looked kinda weird and this normalizes it a little bit.

app/models/notification.rb
```diff
@@ -79,7 +83,22 @@ def unarchive_if_updated
...
     self.attributes = attrs
+    set_author(api_response)
     unarchive_if_updated if unarchive
     save(touch: false) if changed?
   end
+
+  private
+
+  def set_author(api_response)
+    /\/((?:issues|pulls)\/(?<issue_number>\d+))|((?:commits)\/(?<commit_sha>[0-9a-f]{5,40}))\z/ =~ api_response[:subject][:url]
```
tarebyte: That's a helluva regex, but it makes sense 😄

Could we perhaps make that a constant and put it at the top of the model?

```ruby
NOTIFICATION_SUBJECT_TYPE = %r{\/((?:issues|pulls)\/(?<issue_number>\d+))|((?:commits)\/(?<commit_sha>[0-9a-f]{5,40}))\z}

# ....

private

def set_author(api_response)
  NOTIFICATION_SUBJECT_TYPE =~ api_response[:subject][:url]
  # ....
end
```

app/models/notification.rb
```diff
@@ -79,7 +83,22 @@ def unarchive_if_updated
...
+    repo_name = api_response[:repository][:full_name]
+    self.subject_author = if issue_number
+                            issue = user.github_client.issue(repo_name, issue_number)
```
tarebyte: Since the default Octobox doesn't ask for `repo` scope this will raise an `Octokit::NotFound` error when we try grab an issue for a repo we don't have access too.

Just wanted to point it out 😄

andrew:
I think this is great, the only thing that's going to cause headaches is the need for repo scope, at least for the hosted octobox.io version, which I'm only really concerned about the support blowback of people wanting to use it but not give access to code.

app/models/notification.rb
chrisarcand: Gaaah thank you for that. I honestly forgot about it, I have 0 private repos I'm watching and I forget it all the time.

Ugh I don't really know what to do here I guess. @andrew it sounds like you're still too hesitant for octobox.io to ask for repo scope so I'm guessing I should just have this behind a feature toggle. 

I'm hesitant to do that though as it seems like we're sort of reaching a critical mass where to get any more extremely helpful features we just sort of need to bite the bullet and do it one of these times. There's just so much more info given when you can actually query issues themselves.

andrew:
If we go all in and do #4 I'd be very up for requiring repo scope, we can do what I do with Libraries.io and have two oauth apps, one that asks for repo and one that asks for public repo so people can decide what level of access they want to give

chrisarcand:
Closing this in favor of just iterating on it further and making it a part of #367

andrew:
First part of this has been merged in #411 and is available behind the `FETCH_SUBJECT` env var feature flag, docs here: https://github.com/octobox/octobox/blob/master/INSTALLATION.md#downloading-subjects

Still working on getting it available on https://octobox.io

PR #336, opened 2017-04-06
use post request for state changing actions

lalithr95:
Minor change use POST request for the DB state changing actions to prevent CSRF issues.


andrew:
Thanks!

PR #338, opened 2017-04-06
Update Rails to 5.0.2

tarebyte:
If for no other reason than we can get rid of warnings like this one:

```
octobox/octobox/vendor/bundle/ruby/2.4.0/gems/activesupport-5.0.1/lib/active_support/xml_mini.rb:52: warning: constant ::Bignum is deprecated
```


andrew:
🌮

PR #350, opened 2017-05-04
Update to Ruby 2.4.1 (Part Deux)

chrisarcand:
Oops. This stuff should have been included in #349, sorry.

PR #352, opened 2017-05-04
Lock puma to 3.6.2 for now

andrew:
Fixes an issue mentioned by @thiesschneider in https://github.com/octobox/octobox/pull/350#issuecomment-299310166 that affects docker installations, I think related to https://github.com/puma/puma/issues/1284


thiesschneider:
thx, that worked!

PR #364, opened 2017-05-11
Remember login

### LINKED ISSUE #334: Remember login
It would be cool if octobox could remember that you were logged in. It's no huge effort to click the github login button each time but if it would remember I was logged in that would be even better :)

Great work by the way! :tada:

andrew:
Uses a permanent, signed cookie instead of rails session so you stay logged in, fixes #334

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -15,7 +15,7 @@ class ApplicationController < ActionController::Base
   end
 
   def current_user
-    @current_user ||= User.find_by(id: session[:user_id])
+    @current_user ||= User.find_by(id: cookies.permanent.signed[:user_id])
   end
 
   def logged_in?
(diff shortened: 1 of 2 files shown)


acrobat:
Doesn't this cookie need to be http only? As the user id is saved in the cookie? I personally would tag all cookies http only unless you have to access it from javascript

image

> When you tag a cookie with the HttpOnly flag, it tells the browser that this particular cookie should only be accessed by the server. Any attempt to access the cookie from client script is strictly forbidden.

andrew:
The cookie is signed so it can't be tampered with or the user_id read from JavaScript but yeah should probably set it to httponly as well

acrobat:
Yes the userId could not be rad but a xss script could read the content and use that to get logged as that user?

PR #381, opened 2017-05-29
Use Rails 5.1 schema spacing

chrisarcand:
In Rails 5.1, Column arguments are no longer aligned for a given table in `schema.rb`.

For more information, see https://github.com/rails/rails/pull/25675

I'm submitting this commit as it's own change so that we don't have a confusing diff where everything looks different with an actual change hidden therein (like in the thing I'm actually working on...)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -16,37 +16,37 @@ ActiveRecord::Schema.define(version: 20170111185505) do
   enable_extension "plpgsql"
 
   create_table "notifications", force: :cascade do |t|
-    t.integer  "user_id"
-    t.integer  "github_id"
-    t.integer  "repository_id"
-    t.string   "repository_full_name"
-    t.string   "subject_title"
-    t.string   "subject_url"
-    t.string   "subject_type"
-    t.string   "reason"
-    t.boolean  "unread"
-    t.datetime "updated_at",                            null: false
-    t.string   "last_read_at"
-    t.string   "url"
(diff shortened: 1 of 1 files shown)


chrisarcand:
Hold the phone, something ain't right.

db/schema.rb
```diff
@@ -16,37 +16,37 @@
...
+    t.integer "github_id", null: false
+    t.string "access_token", null: false
+    t.string "github_login", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
     t.datetime "last_synced_at"
-    t.string   "personal_access_token"
-    t.integer  "refresh_interval",      default: 0
-    t.index ["access_token"], name: "index_users_on_access_token", unique: true, using: :btree
-    t.index ["github_id"], name: "index_users_on_github_id", unique: true, using: :btree
```
chrisarcand: Ehh, can someone confirm this `using: :btree` removal is no actual change? I think it's the default in Postgres anyway but can't find confirmation that this option was just dropped from the schema in 5.1 or something.

andrew:
Weird, it's not like we're using the schema for anything other than preparing the test db so 👍

chrisarcand:
> Ehh, can someone confirm this using: :btree removal is no actual change?

Randomly stumbled upon it today, here: https://github.com/rails/rails/pull/27981

PR #383, opened 2017-06-01
add support for shift click select of notifications

### LINKED ISSUE #375: Shift-click to select range
I'd like to click to select one item, and then click to select another, non-adjacent item, and thereby select the range (inclusive) bounded by the two items. Currently it only selects the two single items.

n8dzxfzvso

samjacobclift:
❤️ 

multi-select

Fixes #375


samjacobclift:
it maybe actually, I couldn't get `eslint` to run in my editor I'll fix now

Issue #376, opened 2017-05-23
Indicate total number of items


whit537:
I don't see any indication of the total number of items. I end up having to add "Read" and "Unread." Perhaps "Inbox" could have the total number next to it?

![screen shot 2017-05-23 at 10 33 56 am](https://cloud.githubusercontent.com/assets/134455/26359416/67586a46-3fa3-11e7-8393-54f6abac8b27.png)

andrew:
Sounds reasonable!

andrew:
🚢

PR #388, opened 2017-06-04
376 add extra notification counts for inbox, archived and starred fil…

samjacobclift:
…ters

Add some more counts to the sidebar. Fixes this [issue](https://github.com/octobox/octobox/issues/376)
screen shot 2017-06-04 at 20 10 06

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -77,8 +77,6 @@ class NotificationsController < ApplicationController
     @total = scope.count
 
     @inbox_notification    = current_user.notifications.inbox.count
-    @starred_notification  = current_user.notifications.starred.count
-    @archived_notification = current_user.notifications.archived.count
 
     @notifications = scope.newest.page(page).per(per_page)
     @cur_selected = [per_page, @total].min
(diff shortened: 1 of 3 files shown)


andrew:
@samjacobclift The number of unread notifications you have makes me feel very uncomfortable!

samjacobclift:
@andrew you should see the number of emails I have in my inbox!

app/controllers/notifications_controller.rb
```diff
@@ -75,6 +75,11 @@ def index
     check_out_of_bounds(scope)
 
     @total = scope.count
+
+    @inbox_notification    = current_user.notifications.inbox.count
+    @starred_notification  = current_user.notifications.starred.count
+    @archived_notification = current_user.notifications.archived.count
+
```
andrew: I wonder if we can reduce the amount of sql queries here with a group on `:archived`

andrew: Infact, ignore that, can we only show the inbox count and drop the other two? Archived and Starred will only ever go up!

samjacobclift: I can give it a go!

samjacobclift: btw how can a query be saved with that?

samjacobclift:
just the inbox notification count now
screen shot 2017-06-05 at 10 15 56

andrew:
👌

PR #395, opened 2017-06-15
Fix json notification response when zero

sunblaze:
A divide by zero exception occurs when no notifications are returned in json. 

screen shot 2017-06-15 at 2 24 59 pm

Adds a test to verify the pagination attributes of the json response for 1 or more notifications as well as zero notifications. Makes total_pages use the `@per_page` instance variable instead of `@cur_selected`, to avoid the divide by zero issue.


andrew:
Thanks!

PR #396, opened 2017-06-15
Adding shipit.yml so we can deploy with ShipIt!

sunblaze:
https://shipit.shopify.io/shopify/octobox/production already existed, but doesn't seem to be setup properly. Trying to fix that now.


sunblaze:
Sorry, wrong repo

PR #367, opened 2017-05-11
Download subjects

andrew:
Starting to spike out some work on the big one, downloading subjects, also known as #4, starting with the state of the issue/pull request:

screen shot 2017-05-11 at 9 59 48 pm

This opens the door to fixing/closing lots of the frequent requests we have on the issue tracker.

My plan is to store one copy of each subject (Issue, PR, Commit comment etc) and have it referenced from the notification so the data stays as fresh as possible without having to sync everything all the time. 

Current implementation loads issue/pr status on demand (very slow) and only for open source repos, lots still to do here, namely:

- [ ] decide on the data we want to store and be able to query against for each subject type (`RepositoryInvitation`, `Issue`, `PullRequest`, `Commit` and `Release`
- [ ] Cache the relevant subject info into a database table
- [ ] Sync the subject for each new/updated notification
- [ ] Decide on the permissions model and extra opt-in oauth scopes that will be required for this to work for private repos

PR #411, opened 2017-08-12
Download Subjects v2

andrew:
Second pass at #4 as I'm pretty sure @chrisarcand is running https://github.com/octobox/octobox/pull/367 and I didn't want to mangle that branch for him.

This pr adds another table called `Subjects` that caches the `author` for issues, pull requests, releases and commit comments  and `state` (open, closed or merged) for issues and pull requests.

Subjects are only stored once even if many people have a notification for the same issue, linked by the `subject_url` field on `Notification`.

Also allows filtering by `state` from the sidebar menu, screenshot below:

screen shot 2017-08-12 at 10 52 21 pm

This functionality is behind the `FETCH_SUBJECT` feature flag as there are a few incomplete bits, namely:

- [x] Tests
- [x] Index on Subject url field
- [x] Performance improvements for updating subjects when syncing notifications


chrisarcand:
Yeah I have my own branch that's pretty similar to this, but I took a heavy handed approach in that I'm migrating and removing those subject columns off notifications in to the new table. I also did a little (gasp) STI and made out subclasses of subjects for their various specific purposes (commit, PR, etc).

I got a little stuck on something you've mentioned here, the Performance improvements for updating subjects when syncing notifications. I was trying to think of a halfway decent strategy for updating subjects in a reasonable amount of time without just forcing a sync (another API call) on every single new notification, especially since multiple people on the same instance would be subscribed to the same subject and needlessly sync it over and over when a new notification occurs.

Anyway, I'll take a close look at this on Monday and throw some more thoughts on here, see if we can't combine our two paths together and actually get something shipped soon, this issue is long outstanding and it's high time we finished it.

chrisarcand:
@andrew Oh and here, if you wanted to continue this sooner than I can, I pushed the junk I was working on if you care to delve further. Warning: Rough edges.

https://github.com/octobox/octobox/compare/master...chrisarcand:subjects-fork

app/helpers/notifications_helper.rb
```diff
@@ -95,14 +102,27 @@ def no_url_filter_parameters_present
     notification_param_keys.all?{|param| params[param].blank? }
   end
 
-  def notification_icon(subject_type)
+  def notification_icon(subject_type, state = nil)
+    return 'issue-closed' if subject_type == 'Issue' && state == 'closed'
```
chrisarcand: I get why you did this, but feels a bit awkward now that we have more than one decision to make (unlike the other hashes). Maybe, since we only use that constant hash in this spot, we should just put all the logic in here? Like

```ruby
def notification_icon(subject_type, state = nil)
  case subject_type
  when 'Issue'
    if state == 'closed'
      'issue-closed'
    else
      'issue-open'
    end
  when 'PullRequest' then 'git-pull-request'
  when 'Commit' then 'git-commit'
  # ...etc...
  end
end
```

app/models/notification.rb
```diff
@@ -12,6 +15,7 @@ class Notification < ApplicationRecord
                   }
 
   belongs_to :user
+  belongs_to :subject, foreign_key: :subject_url, primary_key: :url, optional: true
```
chrisarcand: Good call 👍

app/models/notification.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 class Notification < ApplicationRecord
+  NOTIFICATION_SUBJECT_TYPE =
+    /\/((?:issues|pulls)\/(?<issue_number>\d+))|((?:commits)\/(?<commit_sha>[0-9a-f]{5,40}))|((?:releases)\/(?<release_id>\d+))\z/
```
chrisarcand: I don't see where this is used; copy/pasted from my branch I think?

app/models/notification.rb
```diff
@@ -79,7 +87,36 @@ def unarchive_if_updated
...
+        author: remote_subject.user.login
+      }
+    when 'Commit', 'Release'
+      remote_subject = download_subject
+      subject_updates = {
+        author: remote_subject.author.login
+      }
+    else
+      {}
+    end
```
chrisarcand: Although I wish this were somehow a part of DownloadService, I do like this much better than my somewhat overengineered STI thing I have going. Simpler.

app/models/notification.rb
```diff
@@ -79,7 +84,36 @@ def unarchive_if_updated
   def update_from_api_response(api_response, unarchive: false)
     attrs = Notification.attributes_from_api_response(api_response)
     self.attributes = attrs
+    update_subject
```
chrisarcand: Although your front end logic is behind the feature flag, this isn't and should be(?)

Though we should do it in conditionally in `update_subject` itself, and fill it what subject stuff we can without needing the permissions for full subject syncing (Always create a subject and fill in with info we currently fetch).

andrew: You're right, I forgot the feature flag here, added it to the `#update_subject` method

app/helpers/notifications_helper.rb
andrew: 18b79f39511f6ae7a40a414de74c5d7d146d9f21e5e36ac7c5e8117b17c6decf

andrew:
I didn't think of moving the other `subject_*` related fields onto `Subject` but that makes sense, the main thing we want to avoid is extra complex joins or indexes needed for filtering, having everything denormalized onto `Notification` makes for a happy postgres database.

I'm going to try some conditional syncing of subjects based on comparing the timestamps between when a notification was created on GitHub and when we last saw the subject updated, ideally for any subject shared by multiple users, the first one that syncs it should pull all the data in and any other users notifications for that same change won't kick off an update because it's already in sync.

andrew:
Pushed a few more performance improvements, the syncing doesn't do any extra http requests than it needs to now, one future optimization we could do is use the graphql api to grab all the notifications in one http request but that feels like we can do that later if needed rather than a must have.

andrew:
The tests are passing now, but that's only because all the new functionality is hidden behind a feature flag, still need to write tests for those bits!

chrisarcand:
@andrew Sadly couldn't finish it today, couldn't get the 2-second no-update tested appropriately. Will pick it up again tomorrow if you don't get to it first.

Hey speaking of which, forgot to comment earlier: What's your thought on the fact that it's two seconds? Trying to come up with a good heuristic for an appropriate number but I'm unsure.

app/models/notification.rb
```diff
@@ -79,7 +84,47 @@ def unarchive_if_updated
...
+
+  def download_subject
+    user.github_client.get(subject_url)
+  end
+
+  def update_subject
+    return unless Octobox.config.fetch_subject
+    if subject
+      # skip syncing if the notification was updated around the same time as subject
+      return if updated_at - subject.updated_at < 2.seconds
```
chrisarcand: Changed the logic here to be a little more intuitive to mean updated within 2 seconds.

chrisarcand:
@andrew Added a few more; I think this is enough but feel free to add more. 

Do you want to address the OAuth stuff here, or merge this and do that separately before you ship to octobox.io?

chrisarcand:
Also just a heads up that due to how we determine to sync notifications, you'll see a mix of authors and non-authors when you flip the switch on this:



The ones without authors are the unread ones. When you read it, the next time you sync the author will show up. I don't really think this is worth addressing but...just FYI.

chrisarcand:
Eh, regrettably I have to report that notifications are no longer syncing when I deployed this branch to my own box. Something's borked.

andrew:
I imagine we'll want a script to backfill subjects for all existing notifications, probably via a rake task

I've not tested the latest changes, will have a look this evening

app/models/notification.rb

andrew:
This is looking pretty good now, going to push `Decide on the permissions model and extra opt-in oauth scopes that will be required for this to work for private repos` into a later pr, just needs some documentation about the feature flag.

andrew:
Added a rake task for syncing missing subjects, @chrisarcand think this is ready to merge?

app/models/notification.rb
```diff
@@ -24,6 +25,11 @@ class Notification < ApplicationRecord
   scope :unread,   ->(unread)       { where(unread: unread) }
   scope :owner,    ->(owner_name)   { where(repository_owner_name: owner_name) }
 
+  scope :state,    ->(state) { joins(:subject).where('subjects.state = ?', state) }
+
+  scope :subjectable, -> { where(subject_type: ['Issue', 'PullRequest', 'Commit', 'Release']) }
```
chrisarcand: Seems weird for some reason. Aren't all notifications 'subjectable' anyway? Or maybe it's more appropriate to say that subjectable means the presence of a `subject_url` 🤔

lib/tasks/tasks.rake
```diff
@@ -9,4 +9,9 @@ namespace :tasks do
       end
     end
   end
+
+  desc "Sync subjects"
+  task sync_subjects: :environment do
+    Notification.subjectable.without_subject.each{|n| n.send(:update_subject) }
```
chrisarcand: * Should probably do this [in batches](http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_in_batches), as I have over 10,000 notifications just myself, so on octobox.io this task might not even work due to all the potential initializations.

* Why `send?`

* Just a thought... Given the number of notifications a person has, this could be pretty bad for someone's rate limit if they have so many notifications. This will be mitigated heavily by one person's notification syncing the subject for everyone else's notification, but 2 seconds might be too fast if the server has too many of them.

Maybe we should have `update_subject` take in a seconds parameter to tweak the distance in time determined to skip syncing a subject. Default it to 2 seconds as it's normally written, but in this rake task specify something quite a bit longer to maximize the potential to skip pointless calls.

app/models/notification.rb
```diff
@@ -79,7 +87,47 @@ def unarchive_if_updated
...
+      when 'Issue', 'PullRequest'
+        remote_subject = download_subject
+        subject.state = remote_subject.merged_at.present? ? 'merged' : remote_subject.state
+        subject.save(touch: false) if subject.changed?
+      end
+    else
+      case subject_type
+      when 'Issue', 'PullRequest'
+        remote_subject = download_subject
+        create_subject({
```
chrisarcand: Realization: Should this code branch be `find_or_create_by`'s or similar? One notification from user A might create a subject and another notification from user B that *doesn't* have that subject associated with it should use the existing subject they share, not create their own (or is that what these meta methods do? I forget, I haven't used them in quite a while).

app/models/notification.rb
andrew: Currently there's one type of subject that doesn't have anything to download `RepositoryInvitation`, which has a subject url like `https://github.com/octobox/octobox/invitations`.

I can imagine GitHub adding more types in the future which is why it's listing the 4 types we know are supported by the the `#update_subject` method.

lib/tasks/tasks.rake
andrew: Good catch, used `find_each` instead.

The 2 second bit isn't for timing or delaying downloads, it's only used when updating existing downloaded subjects to check if the newly synced notification was created around (within 2 seconds) the same time as the time was last saw that the subject was updated, if so we know there's no need to fetch an update.

I'd suspect that the `sync_subjects` rake task would tend to hit peoples rate limits the first time around, the nice thing about using the `without_subject` scope is that it can be ran multiple times and will only download stuff it didn't get the first time around.

app/models/notification.rb
andrew: I did start with a `find_or_create` but the logic so different based on if it's a existing subject or one we need to download from fresh that it was more hassle than it's worth.

The nice thing about using `subject_url` as the foreign key is that every users notification for a given issue/pr/release will reference exactly the same subject, so we'll often have an existing subject in the db which we'll only conditionally need to check to see if it's updated based on that 2 second updated_at comparison.

app/models/notification.rb
chrisarcand: 👍

PR #412, opened 2017-08-17
Changed loading of lib/octobox to fix test coverage

andrew:
I noticed that the test coverage report was marking `lib/octobox` and `lib/octobox/configurator` as having 0% coverage, this moves the require into an initializer so that simplecov sees the file being required.


chrisarcand:
Hah, good catch!

PR #413, opened 2017-08-17
Remove contibutors.json fixture

chrisarcand:
Found this while writing some tests elsewhere; this is a relic from when we hit the API for this instead of just sending to the graphs page.


andrew:
Good catch!

PR #418, opened 2017-08-22
Rescue Octokit errors when downloading a subject

tarebyte:
We just tried to update our internal GitHub version and ran into a snag where we didn't update the scope for the OAuth app and therefore downloading the subject caused a `500`.

This PR catches the two most likely errors from the API, and short circuits the `update_subject` method if there isn't a remote subject.

Since these are private methods I didn't add any tests, but I'm more than happy to if folks think it's needed.

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -96,7 +96,8 @@ class Notification < ApplicationRecord
 
   def download_subject
     user.github_client.get(subject_url)
-  rescue Octokit::Forbidden, Octokit::NotFound
+  rescue Octokit::Forbidden, Octokit::NotFound => e
+    Rails.logger.warn("\n\n\033[32m[#{Time.now}] WARNING -- #{e.message}\033[0m\n\n")
   end
 
   def update_subject
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -96,24 +96,27 @@ class Notification < ApplicationRecord
 
   def download_subject
     user.github_client.get(subject_url)
+  rescue Octokit::Forbidden, Octokit::NotFound
   end
 
   def update_subject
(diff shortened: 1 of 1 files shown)


app/models/notification.rb
```diff
@@ -96,32 +96,34 @@ def update_from_api_response(api_response, unarchive: false)
...
   def download_subject
     user.github_client.get(subject_url)
+  rescue Octokit::Forbidden, Octokit::NotFound
   end
 
   def update_subject
     return unless Octobox.config.fetch_subject
+
+    remote_subject = download_subject
+    return unless remote_subject.present?
```
chrisarcand: Actually I think this should be placed back in the respective clauses; the purpose here was to avoid making a needless API call by the `return if updated_at - subject.updated_at < 2.seconds`, so this is more than a cosmetic change. I'm up for something more elegant if it keeps that functionality though!

app/models/notification.rb
```diff
@@ -96,32 +96,34 @@ def update_from_api_response(api_response, unarchive: false)
 
   def download_subject
     user.github_client.get(subject_url)
+  rescue Octokit::Forbidden, Octokit::NotFound
```
chrisarcand: Hmm is there some way we can flag this so it doesn't get buried when it's what we expected should work? An entry in the logs, perhaps?

app/models/notification.rb

tarebyte: Done in 98179db

app/models/notification.rb
tarebyte: Done in 284f7e4

chrisarcand:
Wooooo, thanks @tarebyte !

andrew:
👌

PR #420, opened 2017-08-25
Dependency and Mocha updates

chrisarcand:
* Updates addressable, mocha, and public_suffix gems
* Removes new mocha deprecations. Passing a block to the mock object constructor is deprecated. Using Object#tap or defining stubs/expectations with an explicit receiver is the correct way to do this moving forward.


andrew:
👌

PR #425, opened 2017-08-29
Clear filters with escape keystroke

### LINKED ISSUE #415: Keyboard shortcut to clear filters
Would be great to be able to clear filters via a keyboard shortcut, say <kbd>Esc</kbd>.

3minus1:
Worked on Issue #415 

Applied filters can be cleared with Esc key.

Fixes #415

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -174,7 +174,7 @@ var shortcuts = {
   191: openModal,         // ?
   190: sync,              // .
   82:  sync,              // r
-  27:  clearFilters
+  27:  clearFilters       // esc
 };
 
 function cursorDown() {
@@ -280,7 +280,7 @@ function autoSync() {
   hasMarkedRows() || sync()
 }
 
-function clearFilters(){
+function clearFilters() {
   Turbolinks.visit("/");
 }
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/application.js
```diff
@@ -173,7 +173,8 @@ var shortcuts = {
   79:  openCurrentLink,   // o
   191: openModal,         // ?
   190: sync,              // .
-  82:  sync               // r
+  82:  sync,              // r
+  27:  clearFilters
```
samjacobclift: I guess you need a '// esc' like the above lines

app/assets/javascripts/application.js
```diff
@@ -279,6 +280,10 @@ function autoSync() {
   hasMarkedRows() || sync()
 }
 
+function clearFilters(){
+  Turbolinks.visit("/");
+}
+
```
samjacobclift: Just a nit we need a space before the '{'  :] does the 'visit' remove the urls query params too?

app/assets/javascripts/application.js
3minus1: Oh, missed that.

3minus1: Fixed.

app/assets/javascripts/application.js
3minus1: Yes it does.

3minus1: And space added 👍

andrew:
Awesome, thanks @3minus1!

PR #427, opened 2017-08-29
Updated Web App Manifest

tkbremnes:
The current Web App Manifest was not really populated with usable data. This commits adds recommended fields, and includes a range of icon sizes to fit a variety of common use cases.

If accessed on a Chromebook, or on an Android device, and then add the application to the drawer or home screen respectively, it'll look behave and work much nicer.

Home screen
home-screen

Splash screen
splash-screen

Task switcher
task-switcher

The "standalone" attribute means that it'll launch without a URL bar.
in-app


andrew:
Sweet, thanks!

PR #430, opened 2017-08-30
Add .dockerignore file

andrew:
TIL about `.dockerignore` file ([docs](https://docs.docker.com/engine/reference/builder/#dockerignore-file)), so I thought I'd add one to octobox to keep the image size nice and small 🐳

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,16 @@
+.git
+.gitignore
+README.md
+
+log
+tmp
+coverage/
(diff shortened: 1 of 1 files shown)


andrew:
According to docker hub, this reduces the build by 1mb ✨  https://hub.docker.com/r/octoboxio/octobox/tags/

chrisarcand:
🎉

PR #431, opened 2017-08-30
Fixed bug where select-all checkbox stays indeterminate

andrew:
Yesterday I noticed that when you've selected one notification then deselect it, the select-all checkbox stays indeterminate, gif example:

octobox

This PR fixes that odd behaviour and makes sure that both the `checked` and `indeterminate` properties on the input are set correctly as `indeterminate` seems to take priority over `checked` if they are both set to true, gif after fixing:

octobox2


chrisarcand:
Good catch!

Issue #310, opened 2017-02-10
Auto refresh fails when leaving a tab open


chrisarcand:
I haven't looked in to how auto refresh is actually implemented, so submitting this without knowing how easy or difficult it would be to fix.

I leave a Chrome tab pinned to Octobox all the time, and I imagine other people do as well. When I leave my machine for an extended period of time, I frequently come back to this in my pinned tab:

![](http://screenshots.chrisarcand.com/permwhsw0.jpg)

This is really irritating as not only do I have to reload the page, but I can't just reload it as the url isn't the actual path required to get me to my inbox. So I frequently trim off the path and access the page again. Seems unimportant, but doing it over and over and over again gets really old.

Note also that I use my own Octobox hosted with Heroku, so maybe my dyno is falling asleep and this doesn't occur on octobox.io or other hosted scenarios that don't do that sort of thing? I would suggest we still address it if possible as I'm sure plenty of people would love to continue hosting this on their hobby tier Heroku without this issue.

davidcelis:
This seems to be happening more frequently, in my experience. I'll have navigated away from my Octobox tab for maybe only five minutes. When I return, I'm greeted with this:

![](https://cdn.davidcel.is/screens/Screen-Shot-2017-08-30-11-31-37.png)

I can navigate back and then re-sync, but I think that something is messing up with syncing while the tab is in the background.

andrew:
How frequently do you have it refreshing? I would guess that the heroku dyno is running out of ram and hitting swap, which slows down the sync and it ends up taking longer than 30 seconds. How much ram is available on your dyno?

davidcelis:
Looks like we have two 2X dynos, each of which has 1GB of RAM. But you're right that RAM usage is very high, and I do see related timeouts:

![](https://cdn.davidcel.is/screens/Screen-Shot-2017-08-30-11-50-15.png)

chrisarcand:
Yeah, I just use a single free dyno for myself and the memory usage tops out with PumaKillerWorker coming in to clean up. I've also noticed it failing more often.

andrew:
Deployed #432, appears to have reduced the amount of time spent making http requests:

![screen shot 2017-08-30 at 9 13 04 pm](https://user-images.githubusercontent.com/1060/29892932-4d89572e-8dc8-11e7-8a9a-ee4de23f16af.png)

I'm guessing you both are running with subject syncing enabled, which I've not put into production yet so I don't have any stats around that.

[...]

PR #432, opened 2017-08-30
Speed up requests to GitHub API

andrew:
Whilst looking into performance improvements for https://github.com/octobox/octobox/issues/310, I noticed that we've got `typhoeus` in the gemfile but aren't using it. 

I added in `faraday_middleware` and a similar config to one I use on https://libraries.io and it appears to improve syncing speed by approx 40%.

When syncing twice from an empty account:

- Before: 3.997s
- After: 2.557s

Adds libcurl as a build dependency, which I believe is always available on macos, so just added it to the Dockerfile.

I also looked into using [oj](https://github.com/ohler55/oj) for parsing json but the result seemed negligible on performance, although may have an effect on memory usage so might still be worth investigating.


andrew:
Going to merge and deploy this to https://octobox.io and watch newrelic to see what kind of difference it makes over the next few hours 🚢

andrew:
Time spent making http requests definitely reduced over the past 30 mins (with lots of people hitting it):

screen shot 2017-08-30 at 9 13 04 pm

PR #434, opened 2017-08-31
Adding more info about using docker

linuxjuggler:
1. A small modifications to the `Dockerfile` so it will include the command instead of adding it to `docker-compose.yml` which will allow the docker image to run outside docker-compose.
2. Added named volumes to postgres.
3. Using postgres-alpine.
4. Added a shared network interface.
5. Updated the Docker installation section to include how to upgrade the container or how to run the containers without cloning the whole code.


andrew:
This is great, thanks @linuxjuggler!

INSTALLATION.md
```diff
@@ -112,6 +116,36 @@ Octobox will be running on [http://localhost:3000](http://localhost:3000).
...
+
+### Production environment
+
+First, Create a network interface 
+
+```bash
+docker network create octobox-network 
+```
+
+First, download and run postgres instance
```
knolleary: Two `First` steps

docker-compose.yml
```diff
@@ -21,4 +22,12 @@ services:
...
-    command: bin/docker-start
+    networks:
+      - internal
+
+volumes:
+  pg_data:
+
+networks:
+  internal:
+    driver: bridge
```
samjacobclift: why do we networks here?

INSTALLATION.md
linuxjuggler: thanks @knolleary I just fixed the typo and pushed it

INSTALLATION.md
```diff
@@ -96,12 +96,16 @@ rails s
...
+### Using Docker Compose
+
 You can use Docker to run Octobox in development.
 
 First, [install Docker](https://docs.docker.com/engine/installation/). If you've got run macOS or Windows, Docker for Mac/Windows makes this really easy.
 
 > If you have Windows Home Edition, you'll need to download and run [Docker Toolbox](https://www.docker.com/products/docker-toolbox).
 
+Second, download the `docker-compose.yml` file from [here](https://raw.githubusercontent.com/octobox/octobox/master/docker-compose.yml)
+
```
samjacobclift: Should we expect that the developer has cloned the repo here or not?

docker-compose.yml
samjacobclift: ohh it to talk to postgres

linuxjuggler: links is going to be deprecated soon, https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/ so the idea is that containers on the same network can communicate directly, as am not a Ruby developer I couldn't figure out why the main container can't communicate with the database even though you can ping it correctly.

INSTALLATION.md
linuxjuggler: depend if he is going to contribute to the development he should clone the repo, if he just wants to use it he can just download the file.

sadly most of the users clone the repo and build the image locally when they just need to download `docker-compose.yml` file and pull the images from docker hub.

docker-compose.yml
```diff
@@ -21,4 +22,12 @@ services:
       - OCTOBOX_DATABASE_USER=postgres
       - OCTOBOX_DATABASE_PASSWORD=development
       - OCTOBOX_DATABASE_HOST=database.service.octobox.internal
```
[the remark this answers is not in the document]
linuxjuggler: am not sure, to be honest, as this is something need to check with the code and am not a ruby guy to know about it.
I'll look into that but maybe later.

docker-compose.yml
linuxjuggler: I was able to remove the links completely so now we only depend on the networks, but we do need to make sure that the database container will be named correctly.

andrew:
Great works @linuxjuggler, thanks for helping improve the docs here!

docker-compose.yml
```diff
@@ -1,18 +1,17 @@
...
+    volumes:
+      - pg_data:/var/lib/postgresql/data
+    networks:
+      - internal
   app:
-    image: octobox
-    build:
-      context: ./
-      dockerfile: Dockerfile
+    image: zaherg/octobox:latest
```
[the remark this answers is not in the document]
karuppiah7890: Hey @andrew, I think there's a change to be made. @linuxjuggler ?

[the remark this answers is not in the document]
andrew: Ah good spot, just pushed a fix for that to master

linuxjuggler:
thanks @andrew for fixing the docker image and using the official one.

cyberhck:
I saw some commits on upstream, so did a git pull, then saw that docker is changed drastically, thank god it didn't break my local instance (I heavily rely on octobox)

PR #439, opened 2017-09-01
Refactor processing notifications

andrew:
The fetching of notifications for new and existing users is definitely one of the slowest parts of the system, the other being marking a number of notifications as read, which I'll be working on tomorrow with @nodunayo, especially with a number of people running #4 in production.

The DownloadService potentially makes a lot of http requests and a high number of sql queries, especially for when someone new signs up, this makes a start at refactoring some of that download logic so we can batch up http requests and sql queries to be able to speed the whole thing up.

First up, I've combined the `process_unread_notifications` and `process_read_notifications` methods into one method called `process_notifications` to focus the next stage of optimization there.

### DIFF
--- a/app/services/download_service.rb
+++ b/app/services/download_service.rb
@@ -49,7 +49,7 @@ class DownloadService
     headers = {cache_control: %w(no-store no-cache)}
     headers[:if_modified_since] = user.last_synced_at.iso8601 if user.last_synced_at.respond_to?(:iso8601)
     notifications = fetch_notifications(params: {headers: headers})
-    process_unread_notifications(notifications)
+    process_notifications(notifications, unarchive: true)
   end
 
   def fetch_read_notifications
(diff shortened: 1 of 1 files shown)


app/services/download_service.rb
```diff
@@ -58,39 +58,24 @@ def fetch_read_notifications
...
-  end
-
-  def process_read_notifications(notifications)
-    return if notifications.blank?
-    existing_notifications = user.notifications.includes(:subject).where(github_id: notifications.map(&:id))
-    notifications.each do |notification|
-      next if notification.unread
-      n = existing_notifications.find{|n| n.github_id == notification.id.to_i }
-      n = user.notifications.includes(:subject).new(github_id: notification.id) if n.nil?
+    notifications.reject{|n| !unarchive && n.unread }.each do |notification|
```
chrisarcand: I really dislike the confusing double negative here ("not unarchive"), but the choice here is more deeply rooted in the Notification API so let's leave it be for now I guess.

chrisarcand:
Nice work!

app/services/download_service.rb
andrew: 🙈

PR #440, opened 2017-09-02
Fix muting and marking as read notifications losing current scope when selecting beyond current page size

andrew:
Turns out that the `Mute` and `Mark as Read` functions were missed when https://github.com/octobox/octobox/pull/308 was added, which looks like it's been causing some performance issues with https://octobox.io where people are inadvertently marking all of their notifications as read, not just the ones in the current view.

@nodunayo and I stumbled across this when we were testing some performance improvements and accidentally made 400+ simultaneous http requests to the GitHub API and got blocked in the process 😬


andrew:
:shipit:

PR #444, opened 2017-09-06
Handle GitHub API failures more elegantly on syncs

### LINKED ISSUE #310: Auto refresh fails when leaving a tab open

chrisarcand:
Closes #310, I'm fairly certain.

Like we do with the sync task, we should be catching API errors from GitHub and reporting on it, not falling apart and responding with a 500 error.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -184,6 +184,12 @@ class NotificationsController < ApplicationController
       format.html { redirect_back fallback_location: root_path }
       format.json { head :ok }
     end
+  rescue Octokit::BadGateway, Octokit::ServiceUnavailable
+    flash[:error] = 'Syncing notifications with GitHub failed. Please try again.'
+    respond_to do |format|
+      format.html { redirect_back fallback_location: root_path }
+      format.json { head :service_unavailable }
+    end
   end
 
   private
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -184,7 +184,8 @@ class NotificationsController < ApplicationController
       format.html { redirect_back fallback_location: root_path }
       format.json { head :ok }
     end
-  rescue Octokit::BadGateway, Octokit::ServiceUnavailable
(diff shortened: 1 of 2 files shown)


test/controllers/notifications_controller_test.rb
```diff
@@ -239,6 +239,23 @@ class NotificationsControllerTest < ActionDispatch::IntegrationTest
     assert_response :ok
   end
 
+  test 'gracefully handles failed user notification syncs' do
+    sign_in_as(@user)
+    User.any_instance.stubs(:sync_notifications).raises(Octokit::BadGateway)
```
chrisarcand: The use of `any_instance` here makes me cringe, but alas it's how we're doing other such specs.

We should refactor the test helper a bit and see about allowing an actual user object to be passed to avoid this but leaving it for now.

app/controllers/notifications_controller.rb
```diff
@@ -184,6 +184,12 @@ def sync
       format.html { redirect_back fallback_location: root_path }
       format.json { head :ok }
     end
+  rescue Octokit::BadGateway, Octokit::ServiceUnavailable
```
tarebyte: What do you think about adding the real error to the server log?

chrisarcand: Losing it from the log is certainly silly, good catch. 👍

andrew:
🚢 on https://octobox.io

PR #449, opened 2017-09-06
Add subject syncing to octobox.io omissions

chrisarcand:
I know some of the larger issues we've kept open pending deploying to octobox.io, but I'd like to add subject syncing and anything else to this list that isn't depoyed on octobox.io so we can actually close the related issues; it'd be a bit nicer to triage what's actually to-be-done in the issue tracker. The example here would be #4 

wdyt @andrew?


andrew:
Makes sense 👍

PR #450, opened 2017-09-07
Use API-given HTML URLs instead of parsing

### LINKED ISSUE #103: incorrect link for releases
I'm using the [hosted version](https://octobox.io) and I noticed that the link given for GitHub releases is incorrect and results in a 404 error.

Examples:
My notification for https://github.com/iuscommunity-pkg/python35u-mod_wsgi/releases/4.5.10-1.ius links to https://github.com/iuscommunity-pkg/python35u-mod_wsgi/releases/4921688.
My notification for https://github.com/iuscommunity-pkg/mariadb101u/releases/10.1.20-1.ius links to https://github.com/iuscommunity-pkg/mariadb101u/releases/4936375.

chrisarcand:
This makes the necessary changes to sync the `html_url` field found on subjects and use that, if possible, instead of doing the hacky API => HTML URL parsing we currently do with notifications only. Note that it's currently backwards compatible with apps not using subject fetching.

In the future, I am hoping we stop kidding ourselves that we can continue to support a non-subject-fetching configuration of the app, what with the plethora of great features subject syncing is getting us and feedback we're getting from it. :stuck_out_tongue_winking_eye: In that event, we'd then remove the hacky parser entirely and just have something for comment anchoring only, as noted in my deprecation comment.

Also note that the backwards compatibility bit enables you to not bother having to go back and re-sync all subjects and notifications, if you don't want to. For example, note the URLs on the two releases here, where 1.5.1 has a subject and 1.5.0 does not:

Fixes #103

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -122,6 +122,7 @@ class Notification < ApplicationRecord
         create_subject({
           state: remote_subject.merged_at.present? ? 'merged' : remote_subject.state,
           author: remote_subject.user.login,
+          html_url: remote_subject.html_url,
           created_at: remote_subject.created_at,
           updated_at: remote_subject.updated_at
         })
@@ -131,6 +132,7 @@ class Notification < ApplicationRecord
 
         create_subject({
           author: remote_subject.author.login,
+          html_url: remote_subject.html_url,
           created_at: remote_subject.created_at,
           updated_at: remote_subject.updated_at
         })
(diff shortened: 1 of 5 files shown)


andrew:
👌

andrew:
Will merge this when I'm at a computer so I remember to deploy with migrations

andrew:
Deployed, not that anyone would notice since it's not enabled yet

PR #455, opened 2017-09-10
Allow API requests past protect_from_forgery

sunblaze:
Allows API requests a method to bypass the CSRF protection. Adding a custom header that disables the protection makes it impossible for a browser to disable the CSRF protection but easy for the application using the API to opt out. Signed cookies are still used for authentication, so the API requests still need to authenticate first and pass the user_id.


andrew:
I'd ask you to update the docs but i don't think there are any docs for the API yet 🙈

Thanks for the patch 👍

PR #457, opened 2017-09-11
Fix #454 - add unread count to window title

### LINKED ISSUE #454: Unread count in title
First of all, I just discovered Octobox and I'm already in love with it, thanks!

It would be awesome if it would show the unread count in the window title - `Octobox (2)` - I think Facebook does this for example. Would be most effective when used with automatic refresh enabled.

roblourens:
By the way, when I'm using Docker, how do I actually update it with my local changes? I can't get it to serve the updated .js file. I'm new to Docker...

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -43,6 +43,9 @@ function moveCursorToClickedRow(event) {
 function updateFavicon() {
   $.get( "/notifications/unread_count", function(data) {
     var unread_count = data["count"];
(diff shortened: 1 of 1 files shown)


andrew:
Thanks for this @roblourens, can you remove the string interpolation? We don't have any transpiler here and I'd like to keep the js as much ES4 as possible for older browser support.

andrew:
Brilliant, thanks!

Issue #458, opened 2017-09-11
Handle API rate limit exceptions gracefully


andrew:
As Octobox.io gets more popular we're starting to trigger exceptions from the GitHub API when people are muting/marking lots of things as read due to the n+1 queries. 

The two exceptions I'm seeing a lot in bugsnag are `Octokit::AbuseDetected` and `Octokit::TooManyRequests`, which should be bubbled up to the user in a similar way to https://github.com/octobox/octobox/pull/444

On a related note, https://github.com/octobox/octobox/pull/441 should help with this, if we start piping everything through curl we can set a max limit on simultaneous requests.

PR #459, opened 2017-09-11
Only mark threads as read if they are currently unread

andrew:
Small optimization to help with https://github.com/octobox/octobox/issues/458, skip marking a notification as read if it's already read, should save a few API calls when users do "mark all as read"


chrisarcand:
👍

Issue #382, opened 2017-05-31
Better Github API error handling


chrisarcand:
We currently just bubble exceptions from things like Octokit, resulting in 500 error pages if, say, Github is having some trouble like they did today. Would be nice to catch these and display a more user friendly flash message or something that Github might be having some trouble.

Specifically we should handle...

* When the GitHub API is down
* When the API key is invalid

andrew:
👍

3minus1:
This could be done by suppressing the console/irb output of the statements where a connection to Octokit is made. We can append a `; nil` to such statements so that the irb doesn't print the return value (an exception like 404 in this case) of those statements. 

Eg: If `notifications = client.notifications(params)` returns an exception for some reason, we can wrap this up in a begin-rescue block like so:

    begin
       notifications = client.notifications(params) ; nil
    rescue
       puts "Something went wrong"
       nil
    end

The exception goes unhandled if we do not use the `; nil` because the irb halts the execution when an exception is returned by the causing statement.

I have tested this in one of my projects and it works. Let me know if I should go ahead and create a PR.

andrew:
@3minus1 I think we actually want to display the reason for the error back to the user on the web page, rather than just in the console

chrisarcand:
#444 handles a small bit of this, but other scenarios like invalid API tokens still need to be done.

ntomas:
Hi! I've submitted a PR #466 with an approach to handle this. Can you tell me if it's a valid one? Thanks

andrew:
@ntomas that's great thanks!

PR #466, opened 2017-09-17
Issue #382 - Handling Github API exceptions

ntomas:
Issue #382 

I tried this approach to deal with the exception in the ApplicationController so there is a way to deal with similar exceptions across all the application. Is this is a valid approach or you prefer other?


andrew:
That's a really neat way to handle this 👍

PR #468, opened 2017-09-18
Add custom bugsnag reporter for users

tarebyte:
From https://docs.bugsnag.com/platforms/ruby/rails/#identifying-users

This will add user identification to an error if the user is logged in.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,7 +6,7 @@ class ApplicationController < ActionController::Base
   helper_method :current_user, :logged_in?
   before_action :authenticate_user!
 
-  before_bugsnag_notify :add_user_info_to_bugsnag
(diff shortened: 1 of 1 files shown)


andrew:
Is the test failure unrelated? Can't tell from my phone. 

I'd guess bugsnag would need to be moved to a runtime dependency if it's not already

chrisarcand:
Huh. 

```
NoMethodError: undefined method `octobox_api_request?' for #<SessionsController:0x0000000003bdf0d0>
    app/controllers/application_controller.rb:5:in `block in <class:ApplicationController>'
```

> I'd guess bugsnag would need to be moved to a runtime dependency if it's not already

Yeah that or just throw an `if Rails.env.production?` on the callback, I don't care.

Issue #327, opened 2017-03-07
Archiving/muting the last message on the last page clears filters


technicalpickles:
Just now I was on the last page of filtered messages by owner and unread (like https://localhost:3000/?technicalpickles=github&page=2&status=true) , and archived the last message. I expected to go to the next to last page with the same filter, but instead I was sent to a (seemingly random) page of notifications without any filter (ie https://localhost:3000/?page=3).

I need to better reproduce this, but I figured I'd write it down in case anyone else has seen it. I think it'd go like:

- choose some filters, making sure there are multiple pages
- go to the last page
- archive or mute the last notification
   - _might_ need to be a single notification on the last page, but I'm not sure

Fryguy:
I am also seeing this, however I had no filters, instead only having the count-per-page set to 100 (I guess that's considered a filter as well?).

andrew:
Fixed by @ntomas in #469

PR #469, opened 2017-09-19
Issue #327 - send all params in the redirect url

ntomas:
Issue #327

Only the page param was being sent. Now it allows all the filters to go through and override the page for the correct page number.


andrew:
Nice fix 👍

PR #473, opened 2017-09-27
Ask for repo scope when subject fetching is enabled

chrisarcand:
Discovered this setting up a new Octobox instance. Confusing to have subject fetching enabled but the application doesn't realize it needs it for new users!

I didn't add a top level fetch_subject_enabled?, I'm not sure if that's actually useful or not, but feel free to add one if you want.


andrew:
Makes sense 👍

PR #474, opened 2017-09-28
Show loading screen on initial notification sync

chrisarcand:
When signing up as a new user, you'll currently stare at a blank screen while the OAuth callback endpoint is getting hit and your notifications are being synced for the first time. For users with a huge number of notifications, this can take a *long* time (multiple minutes) without any feedback to the user at all. It also usually times out and gives a confusing auth error, leaving the user wondering what happened.

These changes make an exception to the initial sync process, instead loading the page and triggering a sync event with a loading screen to assure the user that things are going and it'll take a bit.

Note: I'm *quite* removed from front end work; I know there's a million better ways to probably do this and the Javascript tag in the middle of the markup might be frowned upon. Feedback welcome, though this works just fine for my uses.

Note No. 2: This solution is really just a temporary shim. Syncs should eventually be run in background jobs but I didn't want to open that can of worms right now.

After authorizing the OAuth scopes for a new user, you'll be immediately welcomed to this screen while waiting for your notifications to sync, which stylistically matches the other content we put there:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,7 +3,7 @@ class ApplicationController < ActionController::Base
   API_HEADER = 'X-Octobox-API'
 
   protect_from_forgery with: :exception, unless: -> { octobox_api_request? }
-  helper_method :current_user, :logged_in?
+  helper_method :current_user, :logged_in?, :initial_sync?
   before_action :authenticate_user!
 
   before_bugsnag_notify :add_user_info_to_bugsnag if Rails.env.production?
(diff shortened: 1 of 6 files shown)


chrisarcand:
Gah this is still fine but doesn't fix my issue of users experiencing a terrible onboarding experience. Works great - except on Heroku. TIL Heroku's router purposely errors out the request and emits an H12 error to the log. ([ref](https://devcenter.heroku.com/articles/request-timeout))

andrew:
Looks like a good way to let the users know what's going on, as you say, having a background worker is the next worm-can step, might need a little design love but this is better than it was before.

chrisarcand:
Yeah, whatever, still an improvement. I've brought my erroring out troubles to the above issue regarding background jobs.

PR #477, opened 2017-09-28
Added terms and privacy policy pages for Octobox.io

### LINKED ISSUE #470: Add terms and privacy policy
Add a privacy policy 🙊️

...so that my org admin will approve access request 😁

andrew:
Fixes #470, based on the ones I've done for https://dependencyci.com and Libraries.io, links only visible for https://octobox.io, set via an `OCTOBOX_IO` env var.


config/routes.rb
```diff
@@ -26,6 +26,8 @@
     end
   end
 
+  get '/privacy', to: 'pages#privacy'
+  get '/terms', to: 'pages#terms'
```
chrisarcand: Might as well throw the octobox_io config conditional here, too, as the template isn't customizable.

andrew: Good call 👍

andrew:
🚢

PR #479, opened 2017-09-30
Fix bot author URLs

### LINKED ISSUE #478: user linking to a bot/app account creates an invalid link
octobox 255 2017-09-29 16-20-48

The link is linking to https://github.com/greenkeeper[bot] which doesn't exist, but *should* be linking to https://github.com/apps/greenkeeper

Additionally, albeit minor, but there is no space between the "name" and the "[bot]" part, which is a little strange.

chrisarcand:
Fixes #478

I left the author[bot] look because I think it's sort of nice to see that it's a bot; someone else is welcome to make it look prettier if they like.

### DIFF
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -1,7 +1,24 @@
 class Subject < ApplicationRecord
   has_many :notifications, foreign_key: :subject_url, primary_key: :url
 
+  BOT_AUTHOR_REGEX = /\A(.*)\[bot\]\z/.freeze
+  private_constant :BOT_AUTHOR_REGEX
+
   def author_url
-    "#{Octobox.config.github_domain}/#{author}"
+    "#{Octobox.config.github_domain}#{author_url_path}"
+  end
+
+  private
+
+  def author_url_path
+    if bot_author?
+      "/apps/#{BOT_AUTHOR_REGEX.match(author)[1]}"
+    else
+      "/#{author}"
+    end
+  end
+
+  def bot_author?
+    BOT_AUTHOR_REGEX.match?(author)
   end
 end
--- a/test/models/subject_test.rb
+++ b/test/models/subject_test.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class NotificationTest < ActiveSupport::TestCase
+  test 'author_url returns a user url with users' do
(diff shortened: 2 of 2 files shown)


app/models/subject.rb
```diff
@@ -1,7 +1,20 @@
 class Subject < ApplicationRecord
   has_many :notifications, foreign_key: :subject_url, primary_key: :url
 
+  BOT_AUTHOR_REGEX = /\A(.*)\[bot\]\z/
+  private_constant :BOT_AUTHOR_REGEX
+
   def author_url
-    "#{Octobox.config.github_domain}/#{author}"
+    if bot_author?
+      "https://github.com/apps/#{BOT_AUTHOR_REGEX.match(author)[1]}"
```
chrisarcand: * This assumes that bot usernames directly match GitHub app names every time. I don't know if that's actually true or not.
* Hardcoded here - how does this affect GitHub Enterprise, etc? Can you get apps from other places? I don't know if this assumption is correct either.

@tarebyte Can you confirm any of this?

test/models/subject_test.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class NotificationTest < ActiveSupport::TestCase
+  test 'author_url returns a user url with users' do
+    subject = create(:subject, author: "Fryguy")
+    assert_equal "https://github.com/Fryguy", subject.author_url
+  end
+
+  test 'author_url returns a GitHub apps URL with bots' do
+    subject = create(:subject, author: "greenfield[bot]")
+    assert_equal "https://github.com/apps/greenfield", subject.author_url
```
vitalkanev: Invalid link - gives `HTTP 404 - Not found`. It's seems to be a `greenkeeper`

chrisarcand: Hah, I forgot the name already. Fortunately this is just test so it doesn't really matter ^_^ I'll change it anyway, thanks.

app/models/subject.rb
tarebyte: @chrisarcand at the moment GitHub Apps aren’t available in Enterprise. When it does ship it should point to their own instance, so let’s use configured domain for now.

chrisarcand: Excellent, thanks.

andrew:
Seems like a crappy way for GitHub to shoehorn bot users into their existing API 😑

PR #481, opened 2017-10-02
Don't run fetch_unread_notifications for new users

andrew:
I'm pretty sure that we don't need to make this extra http request for new users as the `#new_user_fetch` method passes `all: true`, so it already will have synced all the notifications for the user, both read and unread.

Opening this to check that I'm not crazy and/or remember how this crazy bit of code actually works!

### DIFF
--- a/app/services/download_service.rb
+++ b/app/services/download_service.rb
@@ -81,6 +81,6 @@ class DownloadService
   def new_user_fetch
     headers = {cache_control: %w(no-store no-cache)}
     notifications = fetch_notifications(params: {all: true, headers: headers})
-    process_notifications(notifications)
+    process_notifications(notifications, unarchive: true)
   end
 end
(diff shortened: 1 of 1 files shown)


andrew:
After looking at the test suite I'm not sure what to believe anymore!

chrisarcand:
> so it already will have synced all the notifications for the user, both read and unread.

I *think* this is true and this change is fine, though I feel some spelunking might be in order to be sure there wasn't something specific.

TBH I don't recall the specifics of why the decision to make two separate requests was made or the different scenarios of how this request(s) change in each scenario, I just remember it was a logical move to do two separate ones in the normal case. Can look back and ponder later if we're still feeling uneasy.

andrew:
3 months later, looking at this with fresh eyes it was pretty easy to see what was wrong and making the test fail

For new users, `process_notifications` was ignoring their read notifications, adding `unarchive: true` to the new_user_fetch sorts it out.

PR #482, opened 2017-10-02
Insert/Upsert all notifications in one query during sync

andrew:
If we require Postgres 9.5 or greater we can make use of the `on duplicate key update` feature (also know as "upsert"), and put all the freshly synced notifications straight into the database in one query, rather than the many that we currently do, especially for new users.

I'm using the [activerecord-import](https://github.com/zdennis/activerecord-import) gem to do the heavy lifting, wiki page on `on duplicate key update` feature here: https://github.com/zdennis/activerecord-import/wiki/On-Duplicate-Key-Update

Couple of test failures atm which I think are related to `updated_at` timestamps not being set correctly


chrisarcand:
Nice, I was considering this as well when pondering about [Heroku request timeouts](https://devcenter.heroku.com/articles/request-timeout). I didn't know that `activerecord-import` provides support for this, so that's nice.

chrisarcand:
> Couple of test failures atm which I think are related to updated_at timestamps not being set correctly

That's correct. I think it's because upserts are always considered 'changes', so the timestamp is updated for every insertion regardless of whether any data fields are actually changed. We're expecting the timestamps to stay the same in the tests.

Still investigating if

1) We even care about that.
2) If we do, how we can avoid updating the timestamp if nothing changed.

Side note, reading the test output is confusing because we have our assertion order backwards (`assert_equal expected, actual` is what it should be...)

Alexkho82:
#482

Alexkho82:
#482

andrew:
This is going to need to be altered slightly to sync subjects and repositories after it's done the notification upsert as new notification records won't have an id before then, so can't be pushed into sidekiq until saved.

andrew:
I'm going to close this, I can't see it ever getting merged in this form, plus things are pretty quick at the moment so we're not in desperate need.

PR #497, opened 2017-11-10
Allow selection of multiple 'reason' filters

### LINKED ISSUE #484: Allow selection of multiple "reason" filters
Specifically there are times I want to be able to hide the "Subscribed" notifications without actually unsubscribing to them forever.

But more generally, it would be good to be able to select more than one "reason" and have a view showing the union of all of those filters e.g. to be able to see `Assign`, `Mention`s and `Review requested`s in one view.

rawburt:
Select multiple "reason" filters and display the union of them.

Fixes #484


rawburt:
@chrisarcand @andrew any thoughts here?

chrisarcand:
Sorry @rawburt I've been really swamped lately and I just haven't gotten to this. One of us will get to this soon, thanks for your patience ❤️

rawburt:
@chrisarcand sounds good chris, thanks!

chrisarcand:
@rawburt Thanks for your patience. 

I noted the joining of multiple reasons in to a single link and tried this out in the UI right away:



The single button looks a bit odd even without the type filter being included and appending a `,` at the beginning; I think in order for this to be added as a feature you'll need to rework it a bit to allow for each reason to be its own filter button, which you can see across filter types:



I can see you're trying to save real estate with multiple filters (many filters would take a lot of room); I think that fixing that is beyond the scope of what you're trying to add here though and would require a larger filter bar rework (perhaps you should be able to drill down by type of filter in that nav bar, something we don't have yet).

rawburt:
@chrisarcand aha! missed that. i'll see what i can do! thanks for the feedback 👌

andrew:
@rawburt I've rebased your branch against master and split out the individual reason filter display as @chrisarcand mentioned, making it ready to :shipit:

PR #498, opened 2017-11-13
Add handler for notifications/sync on GET request

jules2689:
Sometimes when you leave Octobox open for a long time, a request to `POST /notifications/sync` gets interrupted. This, in turn, causes the app to send a `GET` request and 404.

Instead, handle that failure case by allowing `POST` or `GET`. `GET` is basically undocumented, as it's not listed in the `notifications_controller` doc, so the only supported way would remain a `POST` with a fallback on `GET` for this edge case.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,9 +15,10 @@ Rails.application.routes.draw do
     collection do
       post :archive_selected
       post :sync
+      get  :sync
       post :mute_selected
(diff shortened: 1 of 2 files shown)


jules2689:
cc @tarebyte

chrisarcand:
I don't know why we didn't think of this earlier 😅  Thanks @jules2689 !

PR #499, opened 2017-11-13
Mysql and pg

jules2689:
Problem
---
In our infrastructure, we support MySQL not Postgres. While we love using Octobox, using the Postgres datastore puts a restriction on us.

How this solves the problem
---
- To try and solve this, I've added MySQL support via an environment variable `DATABASE=mysql`.
- This defaults to PG still
- It adds a matrix build to Travis to run tests against both systems

WIP to get travis to run the tests


jules2689:
I did not realize the hard dependency on `pg_search`

chrisarcand:
@jules2689 Note I'd welcome any PR that allows for any MySQL-compliant alternatives when using MySQL instead of Postgres! ❤️

jules2689:
@chrisarcand appreciate that. The best I found is https://github.com/makandra/minidusen, which does not seem to give us appropriate compatibility.

MySQL supports full text search in only the latest versions (or so I'm told). So we could introduce the above gem to fall back on, but it won't be as good of an experience.

chrisarcand:
A little support is better than no support, so if you do happen to be running the latest versions of MySQL and only add support for a newish version that's ok if you just want to do that IMO.

jules2689:
Sorry, I wasn't clear. New MySQL stuff supports full-text search, but I'm not aware of any gems that make use of those features yet 😓 https://github.com/makandra/minidusen is basically just fuzzy matching without scoring

chrisarcand:
> New MySQL stuff supports full-text search, but I'm not aware of any gems that make use of those features yet

Oooooh gotcha. Bummer 😕

andrew:
There's always LIKE queries as a fallback 😉

jules2689:
😭 _Fine_, I'll poke through it

app/controllers/notifications_controller.rb
```diff
@@ -192,13 +194,13 @@ def selected_notifications
   end
 
   def current_notifications(scope = notifications_for_presentation)
-    sub_scopes = [:repo, :reason, :type, :unread, :owner, :state]
-    sub_scopes.each do |sub_scope|
-      scope = scope.send(sub_scope, params[sub_scope]) if params[sub_scope].present?
+    [:repo, :reason, :type, :unread, :owner, :state].each do |sub_scope|
+      next unless params[sub_scope].present?
+      type = scope.klass.type_for_attribute(sub_scope.to_s).class
```
jules2689: This is caused by a Rails bug :( It is fixed in 5.2 now though

test/lib/tasks_test.rb
```diff
@@ -7,6 +7,7 @@ class TasksTest < ActiveSupport::TestCase
     travel_to "2016-12-19T19:00:00Z" do
       user = create(:user)
 
+      Rails.application.load_tasks
```
jules2689: Without this, I couldn't get the tests to pass locally :(

jules2689:
@andrew @chrisarcand I got the dual DBs working. I need to clean up the commit history, but WDYT?

db/schema.rb
```diff
@@ -47,7 +44,7 @@
     t.index ["url"], name: "index_subjects_on_url"
   end
 
-  create_table "users", id: :serial, force: :cascade do |t|
+  create_table "users", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4" do |t|
```
jules2689: The schema seems to change whenever I switch back and forth :( I'll try to change it so that this doesnt happen

jules2689:
The schema seems to get changes to it because of a change in Rails 5.1. This issue https://github.com/rails/rails/issues/26209 summarizes it, with this comment: https://github.com/rails/rails/issues/26209#issuecomment-308154650 being what seems to be a preferred solution (?)

chrisarcand:
Cool! I'll look at this soon; there's some things I want to look at a bit more closely but I can say right away with a quick glance: 

* I'm thinking we don't need the `.database` file as well as the env variable; is there a specific reason you wanted to add that? Maybe instead, to stay with all the other settings, you could add the option in [the Configurator](https://github.com/octobox/octobox/blob/master/lib/octobox/configurator.rb)
* With the ability to switch adapters as you've written it, I'd do a `require: false` in the Gemfile for both the db gems and then just require the appropriate one in the adapter logic (don't load both each time)

jules2689:
> I'm thinking we don't need the .database file as well as the env variable

It would suck if I want to develop locally and I would always have to remember the `DATABASE=mysql2` variable. I thought this would be similar to `.ruby-version`

> Maybe instead, to stay with all the other settings, you could add the option in the Configurator

Do you mean the entirety of the `DatabaseConfig`? If so, I looked at that Configurator file just now (didnt know it was there!), but I think adding the DB would make it overloaded. Maybe it would be time to namespace and include, so `Configurator`, `Configurator::Database` etc

> I'd do a require: false in the Gemfile 

:+1: 🤦‍♂️ how didn't I think of that?!

andrew:
We use `dotenv-rails` so you can put `DATABASE=mysql2` in a `.env` file and it will be picked up automatically in development/test

jules2689:
@andrew @chrisarcand I've removed the file support. What are your thoughts on the Configurator worries?

andrew:
FYI I’m off sick at the moment

jules2689:
🏓 @chrisarcand

jules2689:
Travis failed on a remote fetcher error from RubyGems. I have no way to restart it myself it seems :(

INSTALLATION.md
```diff
@@ -25,6 +26,18 @@ in your GitHub settings for Octobox to work.
 * [Downloading subjects](#downloading-subjects)
 
 # Installation
+## Database Selection
+
+Octobox supports a few database adapters. The full list can be found [here](https://github.com/octobox/octobox/blob/master/lib/datbase_config.rb#L2).
```
chrisarcand: Make this a permalink with a SHA attached to it (Hit Y on GitHub while highlighting that same line)

app/controllers/notifications_controller.rb
```diff
@@ -192,13 +194,17 @@ def selected_notifications
...
-    sub_scopes.each do |sub_scope|
-      scope = scope.send(sub_scope, params[sub_scope]) if params[sub_scope].present?
+    [:repo, :reason, :type, :unread, :owner, :state].each do |sub_scope|
+      next unless params[sub_scope].present?
+      # This case is required due to a bug in Rails 5.1
+      # TODO: Rails 5.2 fixes this, so this should be removed when that ships
+      # https://github.com/rails/rails/commit/68fe6b08ee72cc47263e0d2c9ff07f75c4b42761
+      type = scope.klass.type_for_attribute(sub_scope.to_s).class
+      val = scope.klass.type_for_attribute(sub_scope.to_s).cast(params[sub_scope])
+      scope = scope.where(sub_scope => val)
```
chrisarcand: Hah, interesting. 👍

app/models/notification.rb
```diff
@@ -1,15 +1,21 @@
...
 class Notification < ApplicationRecord
-  include PgSearch
-  pg_search_scope :search_by_subject_title,
-                  against: :subject_title,
-                  using: {
-                    tsearch: {
-                      prefix: true,
-                      negation: true,
-                      dictionary: "english"
+  if ActiveRecord::Base.connection.adapter_name == 'PostgreSQL'
```
chrisarcand: I believe this should cause a DB connection to be established on application load (with the class) instead of lazily on the first query as one might expect with ActiveRecord normally.

* That doesn't really matter for us, so I guess this is fine ¯\\\_(ツ)_/¯
* On the other hand, why not just use the nice `DatabaseConfig` you just built instead? 😄

test/lib/tasks_test.rb
chrisarcand: Bummer 😕  I wonder why that is. I have no problems running them locally.

Still this is pretty harmless 👍

jules2689:
> I assume you've been using this branch for Shopify this whole time without issue?

We haven't actually. I've been waiting to use the mysql part of it. I can switch us to using the postgres part in the meanttime

jules2689:
@chrisarcand re- 🏓

INSTALLATION.md
```diff
@@ -28,7 +28,7 @@ in your GitHub settings for Octobox to work.
 # Installation
 ## Database Selection
 
-Octobox supports a few database adapters. The full list can be found [here](https://github.com/octobox/octobox/blob/master/lib/datbase_config.rb#L2).
+Octobox supports a few database adapters. The full list can be found [here](https://github.com/Shopify/octobox/blob/6bd151725678f70aa0adda0f7720cc93fb4b284a/lib/database_config.rb#L2).
```
chrisarcand: Switch to use the octobox repo 😜

jules2689: Woops! Fixed!

chrisarcand:
And you only had to wait 2 months! :trollface: 

Thanks so much for your patience and your work! 💙 

:shipit:

PR #502, opened 2017-11-29
Add icon and colour for security alert notifications

andrew:
Security alerts shipped on GitHub and come with a new notification type: https://github.com/blog/2470-introducing-security-alerts-on-github

Currently looks like this: 

image

This PR makes it look like this: 

screen shot 2017-11-29 at 9 42 48 am

The icon and color are quite striking, but then so are the alert types!


BenJam:
👍

PR #506, opened 2017-12-17
Fix escape key on help modal triggering page reload

hakusaro:
The help modal is a bootstrap modal that binds escape to exiting the modal. This conflicts with the escape keyboard shortcut, and as a result, any selections made are lost during the reload event. This PR stops the escape keyboard shortcut from performing its actions unless the help modal isn't visible.

I just started using Octobox today, and while learning it, I opened the keyboard shortcuts only to find my selection reset during close. Since this is probably one of the first ways people learn how to use shortcuts, I think it's worth fixing.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -289,7 +289,9 @@ function autoSync() {
 }
 
 function clearFilters() {
-  Turbolinks.visit("/");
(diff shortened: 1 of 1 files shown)


hakusaro:
Ugh, I don't actually know how to get this working in JavaScript right now. Everything is a race condition and nothing helps anything.

Issue #507, opened 2017-12-18
Escape key clears filters even if shortcut help modal is open


hakusaro:
Are you experiencing this issue in octobox.io or your own instance?
 - [x] octobox.io
 - [x] My instance

🐞 **Problem**
The escape key currently fires both the dialog close event for bootstrap (closing the keyboard shortcut help menu) and also the clear filters event. If the help dialog is open, it should only close the help dialog.

🎯 **Goal**
The help dialog closes, filters don't refresh. This means that if you're learning the app and need to refresh shortcuts mid-selection, your selection isn't obliterated by reloading the page.

💡 **Possible un-solutions**
Well, I tried to implement this about 10 different ways, most recently in #506.

1. Since the escape event for clearing filters happens "nearly at the same time" as the Bootstrap event, it's virtually impossible to check the state of the modal to see if it's open and cancel taking action. The modal is always "closed" to the keyboard shortcut handler. I've tried checking to see if it's open via a hacky workaround poking directly at bootstrap, and it's always closed.
3. Event handlers on show/hide get fired okay, but the "hide" event handler for the modal gets called before the keyboard shortcut happens, so any state information can't be carried there.
4. Checking `e.target.id` in `enableKeyboardShorcuts()` almost works, but if the modal disappears the target never changes away from the help box until you manually focus the page. This is the best solution because it actually stops the page reload, but then it breaks keyboard shortcuts until you do something to reset `e.target.id`

I don't know enough JavaScript or jQuery to know what the most right solution is. I think maybe doing something to reset the page focus after closing the modal is the best bet -- that would change any future events to have the correct target ID (thus, not the help box)...but I don't know how to do this.

I'm eager to learn what the correct solution to this is too.

lewispb:
I think #510 fixes this @hakusaro, does it work for you?

hakusaro:
@lewispb (I'll assume so :))

andrew:
Deploying #510 to heroku now

PR #510, opened 2017-12-25
Fix escape press closing modal and clearing filters

lewispb:
Addresses:  #507

This PR makes changes to what the escape key does. If a modal is open, it closes it, otherwise it clears filters. Rather than doing both of those things at the same time.

After reading https://stackoverflow.com/questions/16693079/how-to-disable-escape-key-for-twitter-bootstrap-modals/17888680 it was clear that removing `tabindex="-1"` was key to this (pun intended).


andrew:
Thanks!

PR #513, opened 2018-01-06
Add forgotten database_config require

jules2689:
I'm not actually sure how I had this running before 😓 
Currently, the application fails to boot because I forgot to require a lib file.

cc @chrisarcand


jules2689:
It works fine now... it must have been spring affecting me locally. 😢

chrisarcand:
YAIIS (Yet Another Issue Involving Spring) 😄  Cheers

PR #521, opened 2018-01-08
 Explicitly require database_config in application.rb

jules2689:
lib files are only eagerloaded in development. This means that
database_config may not be available to boot. It will be fine, however,
if you first boot the application as spring will cache the database.yml file.

The moment you restart or spring dies, you will hit issues due to implicitly
requiring the file. To fix it, we need to explicitly require lib/database_config.rb

cc @chrisarcand

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,4 +1,5 @@
 require_relative 'boot'
+require_relative "../lib/database_config"
 
 require "rails"
 


jules2689:
supersedes #513 by providing a real explanation and a better commit message

andrew:
LGTM 👍

chrisarcand:
> lib files are only eagerloaded in development. 

From what I gather, this isn't true. `/lib` is set to `eager_load_paths` in `application.rb`, and with eager load paths Rails 5 loads it all right away in production and then disables autoloading.

I don't know a ton about spring as I stopped using it when it just became a constant problem (pun intended), so bear with me...does spring use autoload paths? Is this a problem where it gets eager loaded, spring dies, tries to autoload it, fails? Maybe try adding to autoload paths as well and use `enable_dependency_loading` for spring's sake, as I guess eager loading is by default turned off for the development and test envs? I'm still confused as to why I'm not seeing the issue (I don't use spring).

---

I wrote all that last night and forgot to hit the comment button :) Note I know this change is pretty harmless, I'm mostly just confused as to why it really was an issue still. Thanks @jules2689 !

PR #518, opened 2018-01-07
fix database username env variable

hugomarisco:
closes #517

PR #522, opened 2018-01-10
Fix database username env variable for stable release

Manackkel:
Same fix as #518 when using `image: octoboxio/octobox:latest`


tarebyte:
🍰

PR #523, opened 2018-01-10
Rails supports a DATABASE_URL, handle invalid tokens in notifications sync, change subject_title to text

jules2689:
When a DATABASE URL is set, these values end up being wrong but only the adapter matters. The rest of the values are meaningless after boot, except for adapter which is used to check for search functionality

We only need the adapter to be set. The rest of the values _should_ reflect what ends up being set in the database.yml and should not be reliable past that point (when DATABASE_URL is set). This is similar to how Rails works.

Point form
---
- comes with a caveat. Since Rails doesnt load DB URL nicely into any configs we can't rely on it properly
- so Database Config will be wrong if you use a DB URL env var
- Which means we should _not_ rely on database config past configuring database.yml with the exception of the `is_mysql?` and `is_postgres?` checks

cc @chrisarcand

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -50,6 +50,8 @@ class User < ApplicationRecord
   def sync_notifications
     download_service.download
     Rails.logger.info("\n\n\033[32m[#{Time.now}] INFO -- #{github_login} synced their notifications\033[0m\n\n")
+  rescue Octokit::Unauthorized => e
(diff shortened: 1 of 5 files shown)


jules2689:
Ping @chrisarcand 

This now also has a way to handle invalid access tokens in DownloadService. We had an individual leave and the rake sync task was blowing up because it was using an invalid token

jules2689:
Split to

https://github.com/octobox/octobox/pull/531
https://github.com/octobox/octobox/pull/532
https://github.com/octobox/octobox/pull/533

PR #525, opened 2018-01-15
Kill long running requests after 20s in production

andrew:
If they haven't completed by then, heroku will likely kill the request at the 30s mark

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,4 +54,5 @@ group :production do
   gem 'rails_safe_tasks'
   gem 'bugsnag'
   gem 'puma_worker_killer'
+  gem "rack-timeout"
 end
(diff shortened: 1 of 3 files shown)


chrisarcand:
I don't know if this is wise, or at least in such a general fashion; this puts a presumption that everyone's on Heroku which I know isn't true; short of actually implementing background jobs I've considered moving us off of Heroku just to avoid it.

Find some heuristic to do it only with Heroku/limit it to octobox.io/give a config setting for it?

andrew:
Yeah good point, I'll change it to be configurable via an ENV var and default it to zero, so it's disabled unless otherwise set.

jules2689:
Is this needed anymore with https://github.com/octobox/octobox/pull/603?

andrew:
@jules2689 agreed, closing now

Issue #501, opened 2017-11-15
Columns overflow at large widths


chrisarcand:
When project names are large the columns can't handle the overflow, resulting in:

![](http://screenshots.chrisarcand.com/cpdut.jpg)

hugomarisco:
I can work on this one. Should we ellipsize the project name?

chrisarcand:
I would suggest...

* Ellipsize the PR name, actually, as I've seen PRs across many projects where someone copy/pasted something and it's a paragraph long (I know it's wrapped, though)
* Maybe ellipsize the project name but I'm concerned about making that worthless because of long names in project ecosystems like my example above. I note that we don't strip the organization part from the name IF the owner filter is selected. If I have ManageIQ selected, for example, I don't need `ManageIQ/repo_name` in the list there, I need only `repo_name`. That would save a lot of space and at least allow you to see longer repositories names in filtered views.

Toy with it a bit if you like and see what you think makes the most sense :)

hugomarisco:
I created the PR to use a fixed table layout and ellipsize any column that overflows the predefined width.
I couldn't think of a way to fix this using an auto layout (without involving javascript).

Let me know your thoughts and if there's anything I can do to improve it.

pradyunsg:
This issue can be closed.

PR #366, opened 2017-05-11
Upgrade Bootstrap to 4.0.0

andrew:
screen shot 2017-05-11 at 9 34 06 pm

Looks like it's not the simplest upgrade, don't merge this pr but feel free to poke around and fork/update if you're feeling brave.

PR #526, opened 2018-01-15
fix table layout

hugomarisco:


hugomarisco:
fixes #501

hugomarisco:
hello @chrisarcand I've pushed changes to:

- only apply the fixed layout for non-mobile viewports;
- reduce the width of the repo's name column;

I'm not sure if this will be a good solution for all situations (long issue names vs long repo names). With fixed table layout we need to specify the exact width for all the columns but one and let that one adjust according to the viewport's layout.

A better (but much more complex) solution is use javascript to adjust the widths dynamically or use a div based layout instead of a table (taking advantage of flex layout for instance).

Let me know your thoughts.

andrew:
pinging @BenJam has this might be relevant to what he's doing in https://github.com/octobox/octobox/pull/366

PR #527, opened 2018-01-18
Use url parsing for repo image and dashboard link

osdlge:
Just two small changes to take into account when using this on Github enterprise


tarebyte:
Thanks @osdlge!!

PR #531, opened 2018-02-01
Rails supports a DATABASE_URL, support that properly too

jules2689:
When a DATABASE URL is set, these values end up being wrong but only the adapter matters. The rest of the values are meaningless after boot, except for adapter which is used to check for search functionality

We only need the adapter to be set. The rest of the values _should_ reflect what ends up being set in the database.yml and should not be reliable past that point (when DATABASE_URL is set). This is similar to how Rails works.

Point form
---
- comes with a caveat. Since Rails doesnt load DB URL nicely into any configs we can't rely on it properly
- so Database Config will be wrong if you use a DB URL env var
- Which means we should _not_ rely on database config past configuring database.yml with the exception of the `is_mysql?` and `is_postgres?` checks

cc @chrisarcand

### DIFF
--- a/lib/database_config.rb
+++ b/lib/database_config.rb
@@ -2,8 +2,13 @@ module DatabaseConfig
   SUPPORTED_ADAPTERS = %w(mysql2 postgresql)
 
   class << self
+    # The current adapter being used
+    # Takes into account DATABASE_URL
+    #
     def adapter
-      adapter = if ENV['DATABASE']
+      adapter = if ENV['DATABASE_URL']
+        ENV['DATABASE_URL'].split(":").first
+      elsif ENV['DATABASE']
         ENV['DATABASE']
       else
         'postgresql'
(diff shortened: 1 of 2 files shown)


chrisarcand:
This looks good to me; the code makes sense and I'm ready to merge it, but I'm a little confused about your caveat so I'd like to understand a little more, first.

> comes with a caveat. Since Rails doesnt load DB URL nicely into any configs we can't rely on it properly

Huh? I mean it doesn't keep it anywhere, sure, but it (and now we) takes the DB URL in to account so I don't see the point.

> so Database Config will be wrong if you use a DB URL env var

Before, yes, right? This PR corrects that.

> Which means we should not rely on database config past configuring database.yml with the exception of the is_mysql? and is_postgres? checks

This is the part I'm stuck on; why? It looks like you've done a fine job updating `DatabaseConfig` to take everything in to account, so why is it still bad to rely on `DatabaseConfig` for anything other than the `is_mysql?` and `is_postgres?` checks? Is it just the presumption that *what Rails actually does with the configuration you give it* is a more accurate source of truth than this higher layer which only *instructs* Rails (via `database.yml`) what to do? (aka: Use Rails' API, not our own) 

I can't think of a situation where I'd be legitimately concerned about that differing in some way, but if that's really what you're getting at then that's fine; perhaps you should...

* Add a comment at the top of `DatabaseConfig` saying that it's only intended for configuring `database.yml` and that Rails' mechanisms should be used in actual application code.
* Follow through with the previous point and instead of making an exception for the `is_postgres?` and `is_mysql?` checks, just add these methods to `ApplicationRecord` and simply check which adapter is actually loaded (e.g.: `ActiveRecord::Base.connection.instance_of? ...`). Then there's no exceptions, it's literally just "`DatabaseConfig` is nothing more than a way to encapsulate all the logic one might have interpolated in `database.yml` but in a separate class"

jules2689:
Yea, I can take a look at the dB checks moving to app record. The issue here is that database name is not changed with the env var, and short of parsing the dB URL, I doubt I will be able to easily.

I’ll split the dB config up :)

jules2689:
@chrisarcand 🏓

lib/database_config.rb
```diff
@@ -1,3 +1,11 @@
+##
+#
+# This module is intended *only* to manage database.yml
+# It is not accurate if DATABASE_URL is used, therefore
+# do not use this module outside of database.yml
```
chrisarcand: Still confused by this. `DatabaseConfig` *is* accurate when `DATABASE_URL` is used, no? Isn't that the point of `database_url_or_fallback`? My previous review I talked about how I thought your intention was just to keep Rails as the one source of truth, so this comment is a little misleading if that's the case. Maybe just say 'It is preferred that you use Rails' mechanisms directly in application code and leave this only for database.yml' or similar.

app/models/application_record.rb
```diff
@@ -1,3 +1,17 @@
...
+    # A check to see if MySQL is being used
+    #
+    def is_mysql?
+      DatabaseConfig.adapter.downcase == 'mysql2'
+    end
+
+    # A check to see if Postgres is being used
+    #
+    def is_postgres?
+      DatabaseConfig.adapter.downcase == 'postgresql'
```
chrisarcand: I thought we said these should not be exceptions to the rule with `DatabaseConfig` being used and we should just use Rails? I'd expect to see something like `ActiveRecord::Base.connection.class.name == 'ActiveRecord::ConnectionAdapters::MysqlAdapter'` and similar code instead.

lib/database_config.rb
```diff
@@ -14,7 +22,7 @@ def adapter
         'postgresql'
       end
 
-      unless SUPPORTED_ADAPTERS.include?(adapter)
+      unless DatabaseConfig::SUPPORTED_ADAPTERS.include?(adapter)
```
chrisarcand: ??? I don't care really, but I'm curious, did this not work without fully qualifying or something?

app/models/application_record.rb
jules2689: The class can't be used properly :( Spring/Rails/Magic caches it between tests and makes the test suite really 💩

lib/database_config.rb
jules2689: I think I just made a mistake with the refactor :|

lib/database_config.rb
jules2689: I think I confused myself by not updating my description in the last PR.
Looking back, I was a bad developer and didn't change the PR to reflect the changes we had talked about.

Then since I didn't touch this for a bit, I completely forgot I had made those changes

jules2689:
Ping again... Sorry about the confusion. I removed the last commit :)

chrisarcand:
@jules2689 Ooooh haha ok, everything is clear now 👌 Thanks!

chrisarcand:
Ugh, something's weird with Heroku and these changes cause it to use an unknown ActiveRecord adapter:



Note it's `postgres` instead of the correct `postgresql`

However, note that this is because Heroku specifies DATABASE_URL with just `postgres`:



And I note I guess this is expected behavior: https://devcenter.heroku.com/changelog-items/438

...even though it really should be `postgresql`: https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING

I'm guessing Heroku does something special behind the scenes but I'm not sure how to deal with that yet 😕

PR #533, opened 2018-02-01
Change subject_title from string to text to support longer titles

jules2689:
Subject Title is too short. At varchar(255), I was getting sql errors and exceptions due to the subject_title being too long.

This error went away when I switched to text.

cc @chrisarcand


tarebyte:
I guess technically it could the the length of an issue title which we have as `varbinary(1024)`

jules2689:
I was a bit surprised, it passed by in a log so I didn’t have parameters unfortunately :(

Issue #476, opened 2017-09-28
Sync via background jobs


chrisarcand:
At some point we need to add background jobs for syncing user notifications.

Currently we sync in the request, and depending on the number of notifications or how slow the GitHub API is or how much load is on the Octobox instance, this has a very good chance of timing out when Octobox is deployed on Heroku:

https://devcenter.heroku.com/articles/request-timeout

TL;DR: Heroku cuts off requests longer than 30 seconds. If syncing isn't complete by then, you get an application error (logged as an `H12` error in the application logs).

Using a proper background job queue (ex: Sidekiq) would solve this issue.

[Potentially] related issues:
#310 
https://github.com/octobox/octobox/pull/474
https://github.com/octobox/octobox/issues/475

ntomas:
Hi! I would like to help with this. 

Have a few questions around what we do in the UI. After there's a request to sync we can add a flash message telling we're syncing the notifications. But after that do we do some kind of ping to check if the job is done so we can auto-refresh? Or we have a flag telling those user notifications are being synced and ask the user to refresh after some time? 

Thanks!

chrisarcand:
> But after that do we do some kind of ping to check if the job is done so we can auto-refresh?

I would personally prefer this, yes. Honestly if it behaved the exact same way (currently, sync button spins, page is updated) that'd be lovely.

jules2689:
#543 adds the basic pipeline for this. https://github.com/octobox/octobox/pull/603 adds what is described here.

PR #543, opened 2018-02-19
Install Sidekiq, and configure SyncNotifications background job

tarebyte:
Part 1 of https://github.com/octobox/octobox/issues/476

This PR will _just_ install Sidekiq and create a background job for syncing a users notifications. I was working on the parts for the foreground and the PR was getting rather large so I'm just going to split it up into 2 parts 😄

The other part being how we manage it as a user experience.

### Screenshot of UI changes in https://github.com/octobox/octobox/pull/543/commits/4288285bd0b8bcea0d55a21a123bf93c8e0e0220

screen shot 2018-02-19 at 12 28 27

### Checklist:
- [x] Install and configure [Sidekiq](https://github.com/mperham/sidekiq)
- [x] Create `SyncNotifications` background job
- [x] Install Sidekiq Web UI with some kind of admin gateway
- [x] Add tests
- [x] Get feedback
- [x] Setup redis for all of the different docker configurations
- [ ] Update documentation as needed


config/initializers/01_redis.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+module Octobox
+  REDIS_URL = ENV.fetch("REDIS_URL", "redis://localhost:6379")
+end
```
chrisarcand: Should be part of the [Configurator](https://github.com/octobox/octobox/blob/d76bf25df157ddff096dee938f0728fb595496d2/lib/octobox/configurator.rb) class with all the rest of the configuration; or are you having trouble having that loaded before the Sidekiq initializer?

chrisarcand: Should be part of the [Configurator](https://github.com/octobox/octobox/blob/d76bf25df157ddff096dee938f0728fb595496d2/lib/octobox/configurator.rb) class with all the rest of the configuration; or are you having trouble having that loaded before the Sidekiq initializer?

config/secrets.yml
```diff
@@ -10,21 +10,18 @@
...
   secret_key_base: 6e36f2ef31409ba2199a632ca8437a527407b89bf6d983365fb3e5c0d3b7ad9e47af6565d737a8770faf5e82a7e01a97aac6a653362397d0429d6f590a249515
 
 test:
-  <<: *default
   secret_key_base: 0c04521fa37d6d3452a84bb7b94e5fcccbf1e292b4f3345fd576b2f4c101c2559fa02be9ce3a9e5badd156006569f2074c044f5c1516bf85b2a1c8834a4c466b
 
 # Do not keep production secrets in the repository,
 # instead read values from the environment.
 production:
-  <<: *default
```
chrisarcand: I haven't looked in to the latest secrets stuff in Rails; the churn is becoming unbearable. In any case, I can look in to it if you want these changes but aren't they just changing yet again in 5.2?

tarebyte: Who knows ¯\_(ツ)_/¯ I can undo the change if it's not worth it.

config/initializers/01_redis.rb
tarebyte: Nope, forgot it was a thing 😄 

Done in https://github.com/octobox/octobox/pull/543/commits/b233fee697f25c9c7fdcbe4cf0c62430e38ab0f2

andrew:
Looking good, thanks for tackling this! 💯 

One thing we'll need to do is make sure people running their own instance in docker have a fairly smooth upgrade path.

andrew:
Also, once we've got sidekiq all working, we'll be able to enable subject syncing on https://octobox.io 🎉

andrew:
@tarebyte I've added redis to both docker-compose configs

andrew:
Just deployed this to a test heroku app and it worked great!

I think the only thing left to do before merging this PR is move the ADMIN_GITHUB_IDS to the configurator class

tarebyte:
> I think the only thing left to do before merging this PR is move the ADMIN_GITHUB_IDS to the configurator class

Done in adced95

andrew:
🎉

app/models/user.rb
```diff
@@ -23,6 +23,10 @@ class User < ApplicationRecord
   }
   validate :personal_access_token_validator
 
+  def admin?
+    Octobox.config.github_admin_ids.include?(github_id.to_s)
```
jules2689: Thoughts on `return true if Octobox.config.github_admin_ids.empty?`

andrew: Or if it contains some keyword that means everyone?

jules2689: yea, something along those lines. It's always annoying to find github ids as they don't appear on profiles, etc. And we tend to default to open at Shopify, so it's nice to let anyone see the progress if they want :)

PR #546, opened 2018-02-23
Add API Tokens to User for use with the API

jules2689:
What this does
---
- This adds an api token for use with the JSON API to all users
- It will not generate tokens for existing users, but will instead ask them to click the regenerate button (saves a migration)

cc @chrisarcand @tarebyte for review

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -36,8 +36,12 @@ class UsersController < ApplicationController
   #
   def update
     if current_user.update_attributes(update_user_params)
+      if params[:user][:regenerate_api_token]
+        current_user.regenerate_api_token
+      end
+
       respond_to do |format|
-        format.html { redirect_to root_path }
+        format.html { redirect_back(fallback_location: root_path) }
         format.json { head :ok }
       end
     else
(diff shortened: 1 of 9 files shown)


app/controllers/users_controller.rb
```diff
@@ -36,8 +36,12 @@ def edit
   #
   def update
     if current_user.update_attributes(update_user_params)
+      if params[:user][:regenerate_api_token]
+        current_user.regenerate_api_token
+      end
+
       respond_to do |format|
-        format.html { redirect_to root_path }
+        format.html { redirect_back(fallback_location: root_path) }
```
jules2689: When you click "regenerate token" you probably want to be taken back... I think this is fine, but I'm not sure where else the user settings page is?

chrisarcand: No where else than `/settings`

jules2689: Think it makes sense to just redirect to settings?

jules2689:
Here is the UI ^^

chrisarcand:
Does it make sense to only have an option to regenerate and not reshow a token behind the first time you generate it? (ala GitHub's personal access tokens). 

Granted, the data we deal with here isn't exactly super sensitive (maybe?) and this might be nitpicking but it was a thought I immediately had. Thoughts? @tarebyte ?

jules2689:
I'm fairly indifferent, but I think that the one time show would be a bit more complex as we'd have to include some flag stating that it was shown or not.

PR #547, opened 2018-02-23
Setup API Docs generation and include in header dropdown

jules2689:
What this does
---
- Generate API Docs using Rdoc (uses the existing documentation) into `public/docs` and serve via rails.
- If present, show in the header dropdown
image
- Generates in `Dockerfile`, Installation tells people to generate them if they want them to show up

cc @chrisarcand and @tarebyte for review

Best to wait for #546 so this is useful (and correct 😬)

### DIFF
--- a/API_README.md
+++ b/API_README.md
@@ -1,0 +1,40 @@
+API Documentation
+---
+
+This is the API Documentation for Octobox. With this API, you can access and manage your Github notifications and user profile.
+
+## Authentication
+
+Every user has an API Token that you can see on [your settings page](/settings). Octobox uses standard API Token-based authentication.
+
+To use this authentication, simply send an Authentication header to Octobox:
+
+```
+Authorization Bearer <token>
+```
+
+For example, here is a basic Ruby example to get notifications:
+
+```ruby
+require "net/http"
+require "uri"
+
+base_url = "https://<url>"
+url = URI.parse("#{base_url}/notifications.json")
+
+req = Net::HTTP::Get.new(url.path)
+req.add_field("Authorization", "Bearer #{token}")
+res = Net::HTTP.new(url.host, url.port).start do |http|
+  http.request(req)
+end
+
+puts res.body
+```
+
+## Endpoints
+
+The endpoints are listed by controller down the left hand side of these docs.
+
+## Questions?
+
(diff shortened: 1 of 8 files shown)


API_README.md
```diff
@@ -0,0 +1,40 @@
...
+puts res.body
+```
+
+## Endpoints
+
+The endpoints are listed by controller down the left hand side of these docs.
+
+## Questions?
+
+Please refer to the repository for this app. [You can find the repo here](https://github.com/octobox/octobox).
```
chrisarcand: Technically this should be the source of whatever fork is being run, fetching this link from `Octobox.config.source_repo`, but that might be annoying to interpolate in this markdown I guess.

lib/tasks/docs.rake
```diff
@@ -0,0 +1,14 @@
+require 'rdoc/task'
+
+namespace :api_docs do
+  RDoc::Task.new :generate do |rdoc|
+    rdoc.main = "API_README.md"
+    rdoc.rdoc_files.include(
+      "app/controllers/notifications_controller.rb",
+      "app/controllers/users_controller.rb",
+      "API_README.md",
+    )
```
chrisarcand: Maybe splat this for the controller files at least? Or would the output be funny looking without any docs if they aren't present?

jules2689: I'm not sure what you mean, tbh

API_README.md
jules2689: I could try to inject it - but I don't think Rdoc would play nicely with it :(

lib/tasks/docs.rake
chrisarcand: Sorry I must have been tired when I wrote... I meant [glob](https://ruby-doc.org/core-2.4.2/Dir.html#method-c-glob). `app/controllers/*_controller.rb` instead of forgetting to add a line when we make a new controller.

jules2689: Oh, I chose specific ones because the other controllers actually _dont_ have api endpoints

chrisarcand: Oooh I just thought they'd be ignored if that's the case. Well, nevermind then! 👍

chrisarcand:
This looks good! Thanks @jules2689 !!!

PR #552, opened 2018-03-03
Update deps to fix Tidelist CI failures

tarebyte:


tarebyte:
So looking at the errors from Tidelist, I don't think it was actually the Rails version. I think it might've been miss reading the version and thinking it _might_ have CVE's

/cc @andrew

andrew:
Yeah that’s a weird CVE, I’ve reported it as a bug internally

PR #560, opened 2018-03-19
fix: Respect OCTOBOX_DATABASE_HOST env in docker-start

rhuss:
docker-compose.yml in https://github.com/octobox/octobox/blob/6e37cf176361bbbc7fbe553bed2d74839f12a87b/docker-compose.yml#L26
actually refers to this env variable, so it should
be evaluated in the startup script, too.

### DIFF
--- a/bin/docker-start
+++ b/bin/docker-start
@@ -1,7 +1,8 @@
 #!/bin/sh
 set -e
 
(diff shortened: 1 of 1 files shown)


andrew:
Thanks!

PR #562, opened 2018-03-19
chore: Set g=u for /usr/src/app in Dockerfile

rhuss:
This allows the image to be used on OpenShift as well, as OpenShift
starts containers always with random UIDs.

See https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-specific-guidelines
for the OpenShift recommendations, which don't harm for other
use cases, too.


andrew:
Thanks!

PR #563, opened 2018-03-19
feat: OpenShift support

rhuss:
- Prepare image so that it can be used on OpenShift (same as in #562)
- Add an OpenShift template for easy instantiation of octobox, postgres and redis. This
  is the equivalent to a Docker compose file.

The template has been tested on a full OpenShift cluster (so should work on OpenShift online, too) and
on minishift.

Not sure about the directory location for this template, might be better hidden in some subdirectory.
Also, the resource limits are more or less guessed and can be still optimized.
Several features (fetch_subject, auto sync) can be set as template parameters.

Please let me know, whether you consider to include this PR. In that case, I would also add some documentation and usage examples to installation.md


andrew:
@rhuss thanks for contributing this, happy to merge once you've added the docs.

Looks like the openshift template is missing from your last commit?

rhuss:
oops, right. Let me fix that tomorrow along with some docs.

rhuss:
template added, docs will follow tomorrow.

andrew:
Great work, and the blog post showing how to go about doing too is wonderful, thanks so much!

rhuss:
@andrew you are welcome ;-)

And of course kudos back to you, so many thanks back for creating Octobox ! 
It really saved me from inbox hell, and its one of the rare applications that I use every day ;-) Thank you so much !

PR #572, opened 2018-03-26
chore: Allow 'false' or empty value for FETCH_SUBJECT to disable

rhuss:
config 'fetch_subject' should be only true if FETCH_SUBJECT is set to `true` (according to documentation)

### DIFF
--- a/lib/octobox/configurator.rb
+++ b/lib/octobox/configurator.rb
@@ -24,7 +24,7 @@ module Octobox
     end
 
     def fetch_subject
-      @fetch_subject || ENV['FETCH_SUBJECT'].present?
+      @fetch_subject || (ENV.has_key?('FETCH_SUBJECT') && ENV['FETCH_SUBJECT'].casecmp("true") == 0)
     end
     attr_writer :fetch_subject
 


andrew:
@rhuss it would be good to have a test for this too

rhuss:
agreed, but as said as I'm a Ruby and Rails noob (and actually don't really feel the need to learn Ruby ;-), that's the best I probably can do for now. 

Actually, I probably better would have opened an issue instead. The problem that this PR tries to solve is, that even when you provide a 'FETCH_SUBJECT=false' the feature is still enabled as the check only checks for the existance of the env var, ignoring its value.

Feel free to close the PR if this doesn't make sense to you.

andrew:
@rhuss no probs, I've added some tests and merged to master 👍

chrisarcand:
@rhuss It's "common" within Ruby/Railsland to accept any truthy value when looking for environment vars. I'm not saying it's good, it's just something that's really common, likely due to the weirdness that is the concept of truthiness in Ruby. 

Usually when people are looking to formalize it a bit more you'll see a comparison to an array constant including things like '1', 'true', 'yes', whatever. We can expand on that later. Thanks for finding this! 👍

rhuss:
@chrisarcand I found this when doing the OpenShift template as there is a form field in the template which queries for the value of this env var. It will be always set, with an empty value "" if not given. But typically people enter something like "false" here. Not sure whether the original logic would have worked with an empty value for disabling, but I think now its much cleaner.

Thanks for the fix !

chrisarcand:
Huh, an empty string should have worked but I totally believe you. 

`''.present? #=> false` because `''.blank? #=> true`; these methods are [both from Rails](http://api.rubyonrails.org/classes/Object.html#method-i-blank-3F)

Much cleaner indeed 🙇

PR #575, opened 2018-03-26
Store and use github_id on labels

### LINKED ISSUE #561: Missing notifications and labels
 - [ ] octobox.io
 - [X] My instance

🐞 **Problem**

When switching on `FETCH_SUBJECT=true` then each label appears only exactly once in the overview:

image

In the screenshot above, many issue e.g. carry the label `group/ui` but it is shown only on the first line.

andrew:
WIP fixes #561

As mentioned in https://github.com/octobox/octobox/issues/561#issuecomment-376222137, this separates the remote id from github and the local id.

The main thing I've not tested here is the migration path for users who've been running the existing (broken) label support, I think we'll need to either:

- also look up labels by name+colour to make sure we don't get dupes (may require extra indexes)
- just delete all existing labels during the database migration to add the github_id column, that way everyone just redownloads all labels on subjects next time the subject changes.

Either way we might want a rake task that will update the labels on all subjects for a smooth upgrade path for people who've ran into issues.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -231,7 +231,7 @@ class NotificationsController < ApplicationController
   end
 
   def notifications_for_presentation
-    eager_load_relation = Octobox.config.fetch_subject ? :subject : nil
+    eager_load_relation = Octobox.config.fetch_subject ? {subject: :labels} : nil
     scope = current_user.notifications.includes(eager_load_relation)
 
     if params[:starred].present?
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -11,14 +11,15 @@ class Subject < ApplicationRecord
 
   def update_labels(remote_labels)
     remote_labels.each do |l|
-      label = labels.find_by_id(l.id)
+      label = labels.find_by_github_id(l.id)
       if label.nil?
         labels.create({
-          id: l.id,
+          github_id: l.id,
           color: l.color,
           name: l.name,
         })
       else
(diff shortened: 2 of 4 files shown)


rhuss:
I just did an update of my installation, switching on `FETCH_SUBJECT` on now (so I had no labels previously)

I had some slight issues with SSL cert validation for rubygems.org when building the Docker image, as it seems that rubygems.org just switched its server cert today, signed by another CA than previously.

On a first sight, it looks very good ! I will run it over the day and compare it with the results from octobox.io and also monitor the DB logs. 'will report back later.

rhuss:
Verified over the last few hours:

* that labels appear multiple times
* no duplicate id errors in the DB logs
* no discrepancy between octobox.io and my own installation with `FETCH_SUBJECT=true`
* filtering on labels works as epected

So I would recommend to merge this PR now as it is already much better than before with `FETCH_SUBJECT=true` even when the migration might not be complete.

tbh, for me migration in that case is not that important as (a) you get new labels for all new/updated subjects and (b) for me subjects/notifications are more like transient data which I soon archive anyway so its not so important to me if for some older subjects the labels and extra info is missing (but that might be only me).

Thanks a lot for the quick fix !

andrew:
@rhuss thanks for checking it out, I've added a line to the migration that will delete any existing labels which should avoid getting any duplicates for anyone who's been running master for the past week. 

I also added a little speed up when loading labels in the UI 💯

andrew:
🚢

PR #587, opened 2018-04-04
Make all of notifications controller accessible to JSON api like the docs suggest, add notification state to the json output

jules2689:
It seems we were only showing index. A coworker had a small menu bar script that he wanted to sync from, but we had no way to sync from the app. This PR makes the entire controller API accessible like the docs also suggest.

We also add notification state to the JSON output.

cc @chrisarcand

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 class NotificationsController < ApplicationController
-  skip_before_action :authenticate_user!, only: [:index]
-  before_action :authenticate_index!, only: [:index]
+  skip_before_action :authenticate_user!
+  before_action :authenticate_web_or_api!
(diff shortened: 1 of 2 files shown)


andrew:
@jules2689 good catch, would you like to have commit access to the repo?

jules2689:
@andrew sure, how would that work? I get to hit `merge` myself? or...?

chrisarcand:
@andrew +1, Julian has consistently been a great help to the project for months now and should definitely be added as a collaborator if he'd like :) 👏

jules2689:
Talked a bit with @chrisarcand on slack... Sure :)

> you can also consider us a "large" deployed instance using mysql, subjects, org restriction, etc
> that's often the most benefit I can provide

chrisarcand:
Welcome 🤗

andrew:
Added, welcome to the team @jules2689!

PR #589, opened 2018-04-05
Add sidekiq scheduler for notification jobs

jules2689:
Notifications can not be synced using sidekiq! :tada:
If we make use of `sidekiq-scheduler`, we can get notifications to sync automatically on a schedule.

How is this setup
---
- If you set the env var `OCTOBOX_SIDEKIQ_SCHEDULE_ENABLED`, then it will enable schedules using the default built in schedule.
- I don't want to use the default one? Override the path using `OCTOBOX_SIDEKIQ_SCHEDULE_PATH`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ coverage
 public/assets
 public/docs
 .database
+config/sidekiq_custom_schedule.yml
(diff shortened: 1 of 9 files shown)


jules2689:
ooh. I like that idea. Basically my thought is that the user, using a custom schedule, is likely using a docker container and can copy it into the container. But for other areas, a gitignored file is nice

PR #600, opened 2018-04-11
Fix up incorrect Docker deployment instructions

jimmidyson:
Two issues:

* Shell expands env vars before they are passed to Docker, but they are expanded as empty
* Wrong Postgres username env var


andrew:
Thanks!

PR #603, opened 2018-04-11
User sync in the background

### LINKED ISSUE #476: Sync via background jobs

jules2689:
What this does
---
Converts user syncing to allow background syncing by default from the app. Default syncing will still be foreground for non-UI based flows (API).

Adds a syncing poll endpoint that returns a `423 locked` if the user is syncing still and a `200 OK` otherwise. This is accomplished by storing the most recent job id for a user and checking the status of that job.

Why do we need this?
---
Provides a nicer user experience. Instead of a long sync request, we get a short sync enqueue and then auto-refresh when it's done.

It also reduces issues with timeouts for popular users with a lot of notifications (@rafaelfranca *cough*)

TODO
---
I wanted to mainly get your preliminary thoughts on this pattern, @chrisarcand @tarebyte @andrew, before I dedicate the time to clean it up and add tests.

Fixes #476

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -315,7 +315,7 @@ function refreshOnSync() {
 
 function notify(message, type) {
   var alert_html = [
-    "<div class='alert alert-" + type + " fade in'>",
+    "<div class='alert alert-" + type + " fade show'>",
     "   <button class='close' data-dismiss='alert'>x</button>",
     message,
     "</div>"
@@ -409,8 +409,8 @@ $(document).ready(function() {
 
 // Sync Handling
 $(document).ready(function() {
-  if($(".js-syncing").length){ refreshOnSync() }
-  if($(".js-sync").length){ sync() }
+  if($(".js-is_syncing").length){ refreshOnSync() }
+  if($(".js-start_sync").length){ sync() }
 });
 
 var lastCheckedNotification = null;
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -294,6 +294,35 @@ function openCurrentLink(e) {
   getCurrentRow().find('td.notification-subject .link')[0].click();
 }
 
+function refreshOnSync() {
+  if(!$(".sync .octicon").hasClass("spinning")){
+    $(".sync .octicon").addClass("spinning");
+  }
+
+  jQuery.ajax({'url': "/notifications/syncing.json", data: {}, error: function(xhr, status) {
+      setTimeout(refreshOnSync, 2000)
+    }, success: function(data, status, xhr) {
+      if (data['error'] != null) {
+        $(".sync .octicon").removeClass("spinning");
+        $(".header-flash-messages").empty();
+        notify(data['error'], 'danger')
+      } else {
+        location.reload();
+      }
+    }
+  });
+}
+
+function notify(message, type) {
+  var alert_html = [
+    "<div class='alert alert-" + type + " fade in'>",
+    "   <button class='close' data-dismiss='alert'>x</button>",
+    message,
+    "</div>"
+  ].join("\n");
+  $(".header-flash-messages").append(alert_html);
+}
+
 function sync() {
   $("a.js-sync").click();
 }
@@ -376,6 +405,12 @@ $(document).ready(function() {
     $(".current.js-current").removeClass("current js-current");
     $( this ).find("td").first().addClass("current js-current");
   })
+});
+
+// Sync Handling
+$(document).ready(function() {
+  if($(".js-syncing").length){ refreshOnSync() }
+  if($(".js-sync").length){ sync() }
 });
 
 var lastCheckedNotification = null;
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -201,10 +201,36 @@ class NotificationsController < ApplicationController
   #   HEAD 204
   #
   def sync
-    current_user.sync_notifications
+    if params[:async]
+      current_user.sync_notifications(priority: true)
+      flash[:notice] = "Syncing notifcations in the background. The page will refresh automatically"
+    else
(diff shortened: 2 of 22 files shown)


jules2689:
Okay, I yolo deployed this to shopify's instance and it seems to work.
The test failures are related to failure cases that no longer work the same way AFAICT... but I'll look into those more deeply

jules2689:
I'll assume you're ok with this approach then @andrew?
I'll write some tests soon then!

andrew:
@jules2689 yeah looks much more reasonable than what I was messing around with!

test/controllers/notifications_controller_test.rb
```diff
@@ -284,6 +284,20 @@ class NotificationsControllerTest < ActionDispatch::IntegrationTest
...
+    job_id = @user.reload.sync_job_id
+
+    inline_sidekiq_status do
+      post "/notifications/sync?async=true"
+    end
+
+    assert_response :redirect
+    assert_equal 1, SyncNotificationsWorker.jobs.size
+    assert_not_equal job_id, @user.reload.sync_job_id
+    assert_equal "Syncing notifcations in the background. The page will refresh automatically", flash[:notice]
```
tmlayton: notifications

app/controllers/notifications_controller.rb
```diff
@@ -194,13 +194,39 @@ def star
   #   HEAD 204
   #
   def sync
-    current_user.sync_notifications
+    if params[:async]
+      current_user.sync_notifications(priority: true)
+      flash[:notice] = "Syncing notifcations in the background. The page will refresh automatically"
```
tmlayton: Syncing notifications, missing the second `i`

jules2689:
This has been running for some time in Shopify's production instance. https://github.com/octobox/octobox/pull/603/commits/8872ee45ae7762f96cd146a60fbfb7b8f619cd1a includes fixes that we encountered.

@chrisarcand @andrew can I get some eyes on this so I can ship?

chrisarcand:
YOLO'd this to ManageIQ's instance, and two problems:

* Is it wise to default to background syncing? As is, if someone pulls master and runs this they'll get an error as Redis is assumed to be installed and running now.
* Something's borked; when I try to sync it just spins forever and logs yield a call to `/notifications/syncing.json` over and over and over again forever (actual body content is empty `{}`)

I haven't actually tried looking at the code with this experience in mind; don't have time to try and track it down further but will come back soon, looks promising! Rolled back to master for now.

jules2689:
> Is it wise to default to background syncing? As is, if someone pulls master and runs this they'll get an error as Redis is assumed to be installed and running now.

Hmm, I see what you mean. I think this is a discussion we need to have.

> Something's borked; when I try to sync it just spins forever and logs yield a call to /notifications/syncing.json over and over and over again forever (actual body content is empty {})

This is how it works. We either return a 20x response or 40x error (ok or locked). What I think is happening is that the background job wasn't running for your case.

jules2689:
I'd like to try and get this in @andrew @chrisarcand ... this is blocking Shopify's upgrade to the latest octobox

andrew:
Looking good, only problem left to solve is how to gracefully handle users who upgrade and aren't running a sidekiq worker, and if the default is going to require a sidekiq working going forward we'll need to update all the docs and configs to make that clear to anyone setting it up.

Perhaps we can start by making this opt-in for the basic oauth and then have it default for running it as GitHub App, which pretty much requires sidekiq to handle the webhook requests. 

The GitHub App version is also getting close to the point where it's syncing people's notifications automatically in the background as soon as new activity occurs, so hopefully the sync button will become less important going forward.

jules2689:
> The GitHub App version is also getting close to the point where it's syncing people's notifications automatically in the background as soon as new activity occurs

Maybe it would be better to work towards this instead? That seems far more reliable than this solution. We could migrate to making the sync button alert the user that it's no longer necessary? Then stick a scheduled sync every 10m as a reconciliation job?

andrew:
@jules2689 the sync button will still have a use on notifications from repos that you contribute to that don't have the app installed, which won't really affect Enterprise setups but useful for octobox.io

jules2689:
Ah, I see

jules2689:
It seems to me that we're on the verge of needing sidekiq? It's currently assumed on in the master branch, with sidekiq schedule as an option. Also, given the github app stuff, I assume we're going to need to have sidekiq. I'm inclined to specify that sidekiq is required now?

andrew:
There are going to be a lot of existing installs that don't have a sidekiq worker running, `bin/docker-start` just runs `rails s` and I'd prefer not to break their installations if we can avoid it.

app/assets/javascripts/application.js
```diff
@@ -378,6 +407,12 @@ $(document).ready(function() {
   })
 });
 
+// Sync Handling
+$(document).ready(function() {
```
andrew: Does putting this in `$(document).ready` play nice with turbolinks? If you click a filter whilst it's syncing, does it continue to check?

jules2689: seems to work!

andrew: 👍

andrew:
@jules2689 good plan 👍

jules2689:
ping @andrew

PR #613, opened 2018-04-18
Remove patch that was fixed in Rails 5.2

jules2689:
This cast was needed due to a bug in Rails 5.1.
Rails 5.2 fixes this, so we can remove it when that ships: [the fix](https://github.com/rails/rails/commit/68fe6b08ee72cc47263e0d2c9ff07f75c4b42761)


jules2689:
noppe https://github.com/rails/rails/issues/32624

PR #619, opened 2018-04-24
Encrypt access tokens

jules2689:
Access tokens are sensitive pieces of information capable of accessing a users github account.

We were storing these in plain text, which was noticed because debug logging was logging them in production.

This commit changes all access tokens to be encrypted.

Note: This requires a rotation of all user sessions and requires the users to log back in / re-set their personal access tokens. This is required regardless as the tokens would have been leaking into logs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ ruby '2.5.1'
 
 gem 'rails', '~> 5.2'
 gem 'bootstrap'
+gem "attr_encrypted", "~> 3.0.0"
 gem 'jquery-rails'
 gem 'kaminari'
 gem 'local_time'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,8 +46,10 @@ GEM
       public_suffix (>= 2.0.2, < 4.0)
     arel (9.0.0)
     ast (2.4.0)
-    autoprefixer-rails (9.1.0)
+    attr_encrypted (3.0.3)
+      encryptor (~> 3.0.0)
       execjs
+    autoprefixer-rails (9.1.0)
     bindex (0.5.0)
     bootsnap (1.3.1)
       msgpack (~> 1.0)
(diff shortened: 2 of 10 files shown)


andrew:
What's the user experience like for this when an existing install is upgraded? Do users get logged out or need to start from fresh with GitHub oauth?

Octobox.io isn't storing `personal_access_token` as that feature isn't enabled but is storing `access_token`, we should add those two updated field names to https://github.com/octobox/octobox/blob/master/config/initializers/filter_parameter_logging.rb as well

jules2689:
Their user would still exist, so I think they will just need to go through oauth again. The Access Token is brokered on login, so I think we'd need to rotate the cookie to force a log out.

jules2689:
@andrew is this acceptable?

andrew:
@jules2689 sorry I've been on holiday for a little over a week, I'd like to test this out locally to see what the upgrade user experience is like before merging

andrew:
Finally getting round to merging this, the upgrade experience is a bit painful at the moment.

I have a couple ideas on how we can make it more seamless for end users:

- [ ] If a user's access_token is blank, then redirect them to log back in from any page, to avoid silent syncing errors
- [ ] Provide an example command for generating a 32 character password in the instructions
- [ ] Allow values for `OCTOBOX_ATTRIBUTE_ENCRYPTION_KEY` that are longer than 32 chars
- [ ] Can we fall back to using the rails secret if `OCTOBOX_ATTRIBUTE_ENCRYPTION_KEY` isn't provided?

Also I suspect a lot of users who've installed it via docker are probably running the app in development mode, not sure if that will cause issues for them or lack of security

jules2689:
> Allow values for OCTOBOX_ATTRIBUTE_ENCRYPTION_KEY that are longer than 32 chars

We actually cannot do this. Looking into things more, `attr_encrypted` uses an algorithm which requires exactly a 32 byte key.

andrew:
Deploying this out to https://octobox.io now, thanks for the hard work on this @jules2689 🙌

Issue #615, opened 2018-04-19
Add option of opening GitHub issue in same tab


rhuss:
Are you experiencing this issue in octobox.io or your own instance?
 - [X] octobox.io
 - [X] My instance

🐞 **Feature Request**

Add a boolean user settings which if set, configures octobox to open GitHub issues within the main window, replacing the Octobox UI.

🎯 **Goal** & 💡 **Possible solutions**

I really love working with the awesome octobox keyboard shortcuts exclusively. I also use Octobox within Wavebox (also awesome ;-). However, when I select a notification with ENTER, an new window popups in Wavebox, forcing me to use the mouse.

One could argue, that's a wavebox issue. And that's true, but it would help to open the issue within the Wavebox main window so that one could navigate with keyboard shortcuts, using 'back' to return to Octobox.

Alexkho82:
#615

andrew:
This is probably also relevant to users of Franz: https://github.com/danielbayley/franz-recipe-octobox

andrew:
Just opened #888 to solve this

rhuss:
Cool, let me update the OpenShift template with this option + the sideqik thingy ....

PR #629, opened 2018-05-04
Option for open github issue in new tab

dsalahutdinov:
Closes https://github.com/octobox/octobox/issues/615

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -100,6 +100,6 @@ class UsersController < ApplicationController
       params[:user][:personal_access_token] = nil
     end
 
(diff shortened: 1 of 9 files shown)


andrew:
I think this would probably be better as an application level configuration rather than a user one, i.e. enable it for your own octobox setup via an environment variable

dsalahutdinov:
Hi, @andrew !
Thanks for review. It will definitely better to have application level option. So I will change the code

pradyunsg:
@dsalahutdinov Could you look into this?

andrew:
Closing this in favor of configuring the option via an ENV variable: https://github.com/octobox/octobox/pull/888

Issue #639, opened 2018-05-28
empty redis_url string configuration for scheduled server-side syncs 


chaserx:
Are you experiencing this issue in octobox.io or your own instance?
 - [ ] octobox.io
 - [x] My instance

🐞 **Problem**
I was following the instructions for `Scheduling server-side notification syncs` via Option 2 by including `OCTOBOX_SIDEKIQ_SCHEDULE_ENABLED` in my .env file. Strangely, the app would not boot because of a `invalid uri scheme ''` error that could be traced back to [Line 14 of the sidekiq.rb](https://github.com/octobox/octobox/blob/master/config/initializers/sidekiq.rb#L14) initializer. 

It took me a bit to get it, but I think it's because the [redis_url is nil](https://github.com/mperham/sidekiq/issues/3095#issuecomment-238283509). 

🎯 **Goal**
Using sidekiq-scheduler for scheduled background syncs. 

💡 **Possible solutions**

I have a possible solution that I can post in a PR soon, but I'll describe what I've done below. 

Basically, following the configuration option shown in the sidekiq-scheduler docs seems to work and the error is resolved. 

📋  **Steps to solve the problem**

```shell
redis.service.octobox.internal_1     | 1:M 28 May 23:08:47.771 * The server is now ready to accept connections on port 6379
app_1                                | rake aborted!
app_1                                | ArgumentError: invalid uri scheme ''
app_1                                | /usr/local/bundle/gems/redis-4.0.1/lib/redis/client.rb:415:in `_parse_options'
app_1                                | /usr/local/bundle/gems/redis-4.0.1/lib/redis/client.rb:79:in `initialize'
app_1                                | /usr/local/bundle/gems/redis-4.0.1/lib/redis.rb:38:in `new'
app_1                                | /usr/local/bundle/gems/redis-4.0.1/lib/redis.rb:38:in `initialize'
app_1                                | /usr/local/bundle/gems/sidekiq-5.1.3/lib/sidekiq/redis_connection.rb:55:in `new'
app_1                                | /usr/local/bundle/gems/sidekiq-5.1.3/lib/sidekiq/redis_connection.rb:55:in `build_client'
app_1                                | /usr/local/bundle/gems/sidekiq-5.1.3/lib/sidekiq/redis_connection.rb:34:in `block in create'
app_1                                | /usr/local/bundle/gems/connection_pool-2.2.2/lib/connection_pool/timed_stack.rb:171:in `try_create'
[...]

PR #640, opened 2018-05-29
move scheduler config into the server config block

chaserx:
address #639 

this addresses the problem where you might encounter
an `invalid uri scheme ''` error as sidekiq doesn't
seem to know about the redis_url


andrew:
Thanks!

PR #641, opened 2018-05-31
add last sync to filter bar

### LINKED ISSUE #309: Make the sync situation clearer to users
It isn't entirely obvious that new notifications don't show up on page load.  We should make that clear somehow.

One possibility is to have a flash display the first time somebody loads a page with a last_synced_at over an hour old.

2017-02-07 at 7 55 am

chaserx:
Fixes #309  

Last sync info is hidden on phone sized devices, but not landscape tablets. 

Chrome emulated iPad Pro sizing
screen shot 2018-05-30 at 8 22 06 pm

Chrome emulated iPhone 6/7/8
screen shot 2018-05-30 at 8 20 47 pm

Regular Chrome
screen shot 2018-05-30 at 8 25 50 pm

### DIFF
--- a/app/views/notifications/index.html.erb
+++ b/app/views/notifications/index.html.erb
@@ -44,6 +44,11 @@
           <%= mute_selected_button %>
           <%= mark_read_selected_button %>
           <%= render 'filter-list' %>
+          <% if current_user.last_synced_at %>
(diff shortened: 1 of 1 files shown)


andrew:
Thanks!

PR #666, opened 2018-06-22
Show git commit sha on heroku if available 

andrew:
Requires [dyno-metadata](https://devcenter.heroku.com/articles/dyno-metadata) labs feature enabled for the heroku app, which I've done on https://octobox.io. 

To enable the labs feature, run this command:

```bash
heroku labs:enable runtime-dyno-metadata -a your-heroku-octobox-app-name
```

Also links the sha to the commit on the source repository if present


Alexkho82:
#666

Issue #628, opened 2018-05-02
Notifications stuck on repo no longer access to


ringods:
Are you experiencing this issue in octobox.io or your own instance?
 - [x] octobox.io
 - [ ] My instance

🐞 **Problem**

Two months ago, I switched working for one customer to another. I still had some archived but unread notifications from repos I currently no longer have access to. These notifications are currently stuck. If I want to manipulate these, I get an error mentioning that my github token is invalid. See screenshot.

<img width="1792" alt="octoboxnotifications" src="https://user-images.githubusercontent.com/77923/39510019-8cbdee88-4de8-11e8-8e85-b92d4dfcb1eb.png">

🎯 **Goal**

I would like to keep my Octobox notification list clean.

💡 **Possible solutions**

Allow to remove such notifications from the list completely.

chrisarcand:
Dupe of #591 

Yeah definitely an issue we need to address.

andrew:
This should be fixed now by #670

PR #670, opened 2018-07-02
Swallow Octokit::Forbidden errors when marking notifcations as read

### LINKED ISSUE #591: Can't mark a message as read when I don't have access to the repo anymore
 - [x] octobox.io
 - [ ] My instance

🐞 **Problem**
I have a number of notifications in Octobox from repositories that I no longer have access to (no longer a member of the parent organization).

Attempting to mark any of these as "read" results in this error:

> Your GitHub token seems to be invalid. Please try again.

🎯 **Goal**
I'd like to mark these old notifications as read -- I don't care about them anymore.

💡 **Possible solutions**
Instead of raising an error message, Octobox should allow me to mark the issue as read, or just remove them entirely.

(btw, thanks for a really nice tool!)

andrew:
The user doesn't have access to the repository for that notification anymore so it can't be updated via the GitHub API, but we can still mark it as read.

Fixes #591 and #628

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -64,10 +64,14 @@ class Notification < ApplicationRecord
     user = unread.first.user
     conn = user.github_client.client_without_redirects
     manager = Typhoeus::Hydra.new(max_concurrency: Octobox.config.max_concurrency)
-    conn.in_parallel(manager) do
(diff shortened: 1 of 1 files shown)


Alexkho82:
#670

PR #690, opened 2018-07-16
Rescue Octokit::InternalServerErrors

chrisarcand:
The recent GitHub outage causes Octobox to 500 instead of gracefully fail; we should rescue this exception as well.


andrew:
👍

Issue #528, opened 2018-01-22
Filters + smaller width = row cut off


chrisarcand:
The first row of notifications is cut off when a filter is selected AND the window size reaches the point that the CSS readjusts for mobile. Looks like the filter markup in the bar pushes the bar down over it:

![](http://screenshots.chrisarcand.com/qf6qd.jpg)

rubenlozano:
I can help and fix that at some point during the week, Chris.

rubenlozano:
Working on it 👍

Alexkho82:
#528

existentialmoron:
I inspected this issue. And what's actually happening is a row is partially cut off when window width is below 768px. While, the whole first row in hidden when width is 768px or above. May I be assigned this issue?

andrew:
@existentialmoron feel free to tackle this one 👍

andrew:
I think this has been fixed now 🎉

PR #693, opened 2018-07-19
Fixes: #528. Fixed margin issues with table.

existentialmoron:
See #528 
I have added a media query for less than 768px and adjusted margin for width b/w 768px and 991px.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -86,7 +86,10 @@ html {
           left: 16px;
         }
         > .table {
-          margin-top: 46px;
+          margin-top: 79.5px;
+          @media (max-width: $screen-sm-min) {
+            margin-top: 60.5px;
+          }
         }
       }
     }


andrew:
Thanks!

existentialmoron:
Although this fixes the issue of hidden row, I notice now the table has unnecessarily large margin when the label is not applied. One way I can think of solving is this by adding a class to table or panel when there is a label added. That way, I can add different styling when label is there.

Unfortunately, I am unfamiliar with ruby on rails. @andrew  if you can point me where I need to make changes to add classes when label is added, I would be able to work this out.

andrew:
@existentialmoron ah yeah, I didn't notice that initially.

To add a class if a filter has been applied here's the code: 

```erb
<%= 'filtered' if any_active_filters? %>
```

I think you'll want to add it in here: https://github.com/octobox/octobox/blob/2d1ded72be275f96e81002b32577a2e004c5da45/app/views/notifications/index.html.erb#L54

So it looks something like this:

```html
<table class='table table-hover table-notifications js-table-notifications <%= 'filtered' if any_active_filters? %>' data-refresh-interval=<%= current_user.effective_refresh_interval %>>
```

existentialmoron:
@andrew 
I have made a pull #696 for this.
Kindly review :)

PR #696, opened 2018-07-22
Only increase margin when filters are applied

existentialmoron:


andrew:
Sweet, thanks for the quick fix 🎉

PR #704, opened 2018-07-26
Add bugsnag js notifier

bengourley:
Also makes ga/bugsnag setup more consistent. Loads bugsnag gem in development too as that can be handy sometimes! Let me know if that's not desirable and I'll kick it back to production only 👍


andrew:
Sweet! Don’t worry about the test failure, I think I caused those 😅

PR #711, opened 2018-07-30
Remove rails_safe_tasks gem as the behaviour is built into rails 5.2

andrew:
PR that added the functionality to rails: https://github.com/rails/rails/pull/22967


andrew:
Damn, I think I broke the travis integration

PR #717, opened 2018-08-01
Fix API Docs

lugray:
The X-Octobox-API header is strictly only needed for POST requests in order to bypass forgery protection, but may as well generally be included.

https://github.com/octobox/octobox/blob/ad55b1f4d76ac5b57d8300dca921dedb7b9f8b51/app/controllers/application_controller.rb#L5


andrew:
Ah good catch! 👌

PR #734, opened 2018-08-02
Clarify the Heroku deploy button and OAuth

### LINKED ISSUE #718: Octobox Heroku Deploy OAuth Question
 - [ ] octobox.io
 - [x] My instance

🐞 **Problem**

I seem to be having a bit of a chicken-and-egg problem where to deploy to Heroku I need a github OAuth client id/secret, and to create a github OAuth token I need the app homepage url and the github callback url.

image

image

🎯 **Goal**

Get updated documentation on deploying through the deploy button on Heroku, particularly I couldn't find the link to creating a new oauth token (https://github.com/settings/applications), and I am not sure how you solve the url being required on both ends and needing the token to deploy and needing the deployed url to make the token.

💡 **Possible solutions**

chrisarcand:
Closes #718

Adds a bit more about how to fill out your callback URL when registering an OAuth app before you've actually provisioned on Heroku. Attempting to be a bit more helpful without duplicating specifics about the authorization URL found in the local installation instructions, which is mentioned as the next step to look at if someone's still having issues.

@audiolion Does this clarify things a bit for you?

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -47,9 +47,13 @@ You can host your own instance of Octobox using Heroku.
 
 [![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy?template=https://github.com/octobox/octobox)
 
-Heroku will ask you to provide OAuth client ID and secret, which you can create
-on GitHub. When creating the OAuth application, make sure you enable the
-`notifications` scope on it (you will also need the `read:org` scope if you enable restricted access).
+Heroku will ask you to provide an OAuth client ID and secret, which you can get by
+[registering a new OAuth application on GitHub](https://github.com/settings/applications/new)]. When creating the OAuth application:
+
+* Make sure you enable the `notifications` scope on it (you will also need the `read:org` scope if you enable restricted access).
+* You can provide Homepage and Authorization URLs by using the Heroku app name you choose. By default, a Heroku app is available at its Heroku domain, which has the form `[name of app].herokuapp.com`.
+  The callback url would then be `[name of app].herokuapp.com/auth/github/callback`.
+
 For more help with setting up an OAuth application on GitHub, see below.
 
 ## Deployment to OpenShift Online


audiolion:
bitmoji

Yeah that helps! So the callback and homepage are the same? I thought based on the local docs it might be `[appname].herokuapp.com/auth/github/callback`

andrew:
@audiolion you’re right, the callback url should go to `[appname].herokuapp.com/auth/github/callback`

audiolion:
> @audiolion you’re right, the callback url should go to [appname].herokuapp.com/auth/github/callback

I don't see any reason to not explicitly include that in the docs? Maybe I am reading the docs wrong but I went to "Installation Guide" and for Heroku it is just the button, it doesn't say how to fill in the homepage url or callback url, only reason I picked that up was looking at the local install guide where it mentioned it. I normally wouldn't have looked there if I was doing just a Heroku install but I ran into this issue so started reading the whole document

docs/INSTALLATION.md
```diff
@@ -47,9 +47,12 @@ You can host your own instance of Octobox using Heroku.
...
 [![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy?template=https://github.com/octobox/octobox)
 
-Heroku will ask you to provide OAuth client ID and secret, which you can create
-on GitHub. When creating the OAuth application, make sure you enable the
-`notifications` scope on it (you will also need the `read:org` scope if you enable restricted access).
+Heroku will ask you to provide an OAuth client ID and secret, which you can get by
+[registering a new OAuth application on GitHub](https://github.com/settings/applications/new)]. When creating the OAuth application:
+
+* Make sure you enable the `notifications` scope on it (you will also need the `read:org` scope if you enable restricted access).
+* You can provide Homepage and Authorization URLs by using the Heroku app name you choose. By default, a Heroku app is available at its Heroku domain, which has the form `[name of app].herokuapp.com`.
```
audiolion: Suggested edit:
```
* You can provide Homepage and Authorization URLs by using the Heroku app name you choose. By default, a Heroku app is available at its Heroku domain, which has the form `[name of app].herokuapp.com`. The callback url would then be `name of app].herokuapp.com/auth/github/callback`

chrisarcand: Cool, amended!

PR #736, opened 2018-08-03
Add db index to subject_url on notifications

andrew:
Massively speeds up finding all the notifications for a given subject on octobox.io


andrew:
Just to be clear, it isn't slowing down any queries in production right now, other that ones I'm experimenting with adding.

andrew:
From 800ms to 1.5ms 🚀

chrisarcand:
Ahh I thought we were already doing a bunch of lookups via subject_url already. In any case...great!

Issue #472, opened 2017-09-27
Support for optionally enabling downloading subjects


andrew:
The last piece of the puzzle to enable #4 on https://octobox.io is to optionally allow users to give `repo` scope access to Octobox so it can sync the status for issues and pull requests on private repositories. 

In practise what this means is configuring two different oauth apps and allowing the user to choose which one they login with, which is what I've done on both https://libraries.io and https://dependencyci.com, works well given the current state of the GitHub oauth API scopes.

This also means that a user can potentially be logged in via both, once they upgrade their access, so they will have two different token fields on their database record and the one with the higher access will be used if present.

Semi-related to https://github.com/octobox/octobox/issues/250, would certainly allow configuring of one Octobox instance to talk to Github.com and a GitHub Enterprise install at the same time if done right.

chrisarcand:
Dupe of https://github.com/octobox/octobox/issues/417 but this one is more thorough, so closing that one.

andrew:
Sorry, I'm the worse 🙈  working backwards through the issues list to close out old ones, forgetting what things have been opened already

yonaskolb:
Are there any updates on this?

tsupertramp:
@andrew - Wouldn't it make sense to remove the experimental flag (FETCH_SUBJECT=true) and make this default for all users, instead of a complex two way system? - I could PR this and add some text why this is happening for users where scope has to be upgraded.

The second option would be allow upgrading the scope in the settings of the user maybe, I would prefer option one.

Greetings, Thomas

andrew:
@tsupertramp the main thing holding back switching the default on is the github permissions model for oauth apps, for anyone with access to private repos, they need to give a lot of access to code as well as issues and pull requests often to multiple orgs, I'm investigating getting it working as a GitHub App to allow for fine-grain permissions then we can definitely have FETCH_SUBJECT default to true

andrew:
I'm actively working on this now, subject syncing will be enabled for any repos that have an Octobox GitHub app installed, so orgs can opt-in to this behaviour rather than users having to agree to `repo` scope, hoping to have something to show in the next week or so

andrew:
@tsupertramp I've got this working (but hidden) on https://octobox.io now, let me know if you'd like to test it out.

tsupertramp:
Hey @andrew - Totally!

andrew:
[...]

PR #737, opened 2018-08-03
Subject -> notifications -> users

andrew:
I'm working towards being able to sync subjects in the background (sidekiq) via webhooks, related to https://github.com/octobox/octobox/issues/472

This means, if we see a subject has been updated, we can proactively sync all subscribed users (one's who've been notified about that subject before), which will speed up notification and subject syncing as most of the updates will already have been downloaded.

It also has a nice multiplying effect for installations with lots of users, like octobox.io, the more active users syncing notifications, the more up-to-date related users will be.

Dependent on #736 otherwise the performance is terrible!

### DIFF
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -28,6 +28,10 @@ class Subject < ApplicationRecord
     end
   end
 
+  def sync_involved_users
+    user_ids.each { |user_id| SyncNotificationsWorker.perform_async(user_id) }
+  end
+
   private
 
   def author_url_path
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -1,6 +1,7 @@
 class Subject < ApplicationRecord
   has_many :notifications, foreign_key: :subject_url, primary_key: :url
   has_many :labels
+  has_many :users, through: :notifications
 
   BOT_AUTHOR_REGEX = /\A(.*)\[bot\]\z/.freeze
   private_constant :BOT_AUTHOR_REGEX


chrisarcand:
Pondering...wouldn't you want the subject to have the user update only the particular notification/subject in question, instead of having any subject call a full sync for the user (which will be notifications call(s) + 1 for each subject)?

andrew:
@chrisarcand subjects are shared between all users on an installation, so that'll be up-to-date for everyone as soon as the subject gets updated.

My thinking was, this would quickly catch up with the global activity feed of GitHub, basically everyone in sync, then any time there's an update, it syncs users who're subscribed, which in theory would only do the 1 http request (per user) to update the notification records for that subject.

In practice there would be new subjects as well that wouldn't have everyone's notifications sync'd down yet.

Could update only just the existing notifications, but that'd still be one http request per user.

The thing to make sure is that it doesn't get into a cycle of calling `sync_involved_users` which downloads some other subjects and ends up calling `sync_involved_users` over and over.

andrew:
Also investigating using the v4 graphql API to pull all the updated subjects at once when a user sync's which should reduce the n+1 calls for syncing further subjects

chrisarcand:
> subjects are shared between all users on an installation,

Ahhh yes I forgot about that shortcut, good call.

> Also investigating using the v4 graphql API to pull all the updated subjects at once when a user sync's which should reduce the n+1 calls for syncing further subjects

Yeah the more we've been chatting in various issues the more I'm interested in doing a little V4 API spike, might start in on that soon.

PR #743, opened 2018-08-07
Move the release note generation code into it's own class

andrew:
Mostly to stop the tests complaining about constants already being defined.

Only functional change is a little bug fix changing `2&>1 ` to `2>&1` to avoid creating a random file called `1`


jules2689:
speaking of release notes...

andrew:
🎉

PR #745, opened 2018-08-07
Always show read/unread sidebar menu items

### LINKED ISSUE #377: "unread" pane disappears when i have no unreads
I want https://octobox.io/?unread=true to be my homepage, and I really almost never want to look at any other page. However, the instant I achieve my goal of Notification Zero, it disappears from the left nav, and clicking "refreshing" dumps me on the inbox.

Can this just permanently stick around?

andrew:
Shows a zero count if there are no unread messages.

Fixes #377 

Looks like this:

screen shot 2018-08-07 at 10 47 54

### DIFF
--- a/app/views/notifications/_sidebar.html.erb
+++ b/app/views/notifications/_sidebar.html.erb
@@ -29,17 +29,14 @@
 
   <%= menu_separator %>
 
-  <% @unread_notifications.each do |unread_notification, count| %>
-    <%= filter_link :unread, unread_notification, count do %>
-      <% if unread_notification %>
-        <%= octicon 'mail', height: 16, class: 'sidebar-icon star-active' %>
-        Unread
-      <% else %>
-        <%= octicon 'mail-read', height: 16, class: 'sidebar-icon star-active' %>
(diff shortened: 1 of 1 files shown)


app/views/notifications/_sidebar.html.erb
```diff
@@ -29,8 +29,8 @@
 
   <%= menu_separator %>
 
-  <% @unread_notifications.each do |unread_notification, count| %>
-    <%= filter_link :unread, unread_notification, count do %>
+  <% [false, true].each do |unread_notification| %>
```
jules2689: It's a bit counter-intuitive to be `unread_notification` but be for read and unread.
I'm also wondering if the loop is more complex than it needs to be. Is the code duplication here better than having the loop and trying to figure out what `false` and `true` is?

andrew: Agreed, thanks for the second pair of 👀

andrew: Force pushed the simpler version 👍

PR #748, opened 2018-08-07
Fix/721 add momentum scrolling on ios devices

samjacobclift:
Enabled in devices of medium size and under

1. sidebar
2. issue list

Note [this](https://stackoverflow.com/questions/30102792/css-media-query-target-only-ios-devices) seemed to work, so could perhaps we can use this if we are not happy targeting by screen size 😄


andrew:
Yep, that's what I would have done, thanks!!! 👍

PR #750, opened 2018-08-07
Set emoji height in labels to same height as text

andrew:
Labels with emojis are quite a bit taller that ones without, also fixed at 18px, this pr makes them the same size as the font which shrinks them by default and allows them to expand if you increase the font size.

Before:

screen shot 2018-08-07 at 16 44 19

After:

screen shot 2018-08-07 at 16 49 42


app/assets/stylesheets/application.scss
```diff
@@ -643,3 +643,7 @@ td.keys {
 .settings-sha {
   margin: 35px 0 10px;
 }
+
+.emoji{
+  height: 1em;
+}
```
samjacobclift: just a small change but I could we set a width here, I think if we do that the browser will assign the correct space for the emoji during page load i.e. stop a little page shift

andrew: Good call, just added that 👍

andrew:
ship-it-squirrel

PR #752, opened 2018-08-07
Triage view

andrew:
Mentioned this to @milesmcbain on [twitter](https://twitter.com/teabass/status/1026838963615621121), so I thought I'd open it up for discussion.

The idea behind triage view is all the things that have yet to be dealt with, instead of just working directly out of your inbox. 

screen shot 2018-08-07 at 17 49 38

It's a separate constant menu item on the right hand size that lists notification threads with subjects  (i.e. pull requests and issues) that are "open", regardless of their archived state, across everything they are involved in.

It needs `FETCH_SUBJECT` enabled on the install to work, so won't be available on https://octobox.io just yet, but I'm working on making that a possibility within the next couple weeks.

This list is still filterable in all the usual ways:

- type
- reason
- org
- repo
- label
- read/unread

screen shot 2018-08-07 at 17 42 52

Would love to hear what other people's workflows are when triaging issues/pull requests and what's needed to make this really useful.

Some things that are likely already missing:

- Show issues/pull requests that are open on repos you're responsible for that haven't generated any notifications yet
- Filter by "no label" for things that haven't been labelled up yet (where having a label may suggest it's been triaged by someone already)
- Ability to filter out issues/prs that you can't do anything about, most likely because you happened to get mentioned on someone else's repo, perhaps one solution to this is only show threads/subjects that  you are subscribed to, so hitting "mute" would remove them from this view.
- Sorting options (newest, oldest, comment count etc)

Issue #742, opened 2018-08-06
Advanced search query syntax


andrew:
Example:

`unread:true type:issue reason:mention owner:octobox repo:octobox/octobox label:bug free text`

Potential ruby gems we can use for parsing: https://github.com/marcelocf/searrrch

Other potential filters we can support:

- author
- assignee
- subject
- body (#709)
- comments (#709)
- date
- inbox/archive/starred
- team

This is also the first step in enabling custom filters, where if a notification comes in an matches a custom saved search then perform some action, useful for #8, #217 and #599

andrew:
We can also build a drop-down advanced search UI similar to gmails which constructs the query and aids in discover of available filters:

![screen shot 2018-08-06 at 15 30 38](https://user-images.githubusercontent.com/1060/43722698-c9a8bd3c-998d-11e8-8c3e-aa9f2b46ed0e.png)

andrew:
Another follow up will be allowing users to "pin" search queries to the sidebar for easy access (and counts)

pradyunsg:
Ideas for the search bar:

- Use it to replace the filters bar.

- Visually aligning the search bar with the actual table (again, idea from Gmail; done on Octobox by injecting a CSS rule)
  <img width="403" alt="screen shot 2018-08-16 at 3 25 19 pm" src="https://user-images.githubusercontent.com/3275593/44202076-9ef00180-a168-11e8-8472-5dca7d49966d.png">
  ```css
  /* Horrible Hack for aligning search */
  #search {
    position: absolute;
    left: 250px;
  }
  .navbar-brand svg.octobox-icon {
    position: absolute;
    top: 1rem;
    left: 7.5rem;
  }
  ```

- Keep history of searches in the LocalStorage of the user and allow them to remove them (GitLab does something like this IIRC, even Gmail)

andrew:
We should definitely pop the search history item out into it's own issue, I've been meaning to jot that down somewhere.

I also like the idea of bumping the search box across so it aligns with the notifications table, at least on large screens.

andrew:
Copying more elements of the gmail UI just for fun, almost completely unusable but looks nice: 
![screen shot 2018-08-16 at 12 05 02](https://user-images.githubusercontent.com/1060/44204975-b8835000-a14c-11e8-8ab9-c950d40bd737.png)

pradyunsg:
> Copying more elements of the gmail UI just for fun, almost completely unusable but looks nice:

[...]

PR #756, opened 2018-08-08
Add a filter for unlabelled issues and pull requests

andrew:
As mentioned in #752, being able to triage things by focusing on items that haven't been labelled yet is very useful.

Only works if you have subject syncing enabled (`FETCH_SUBJECT`), so won't work on https://octobox.io just yet.

screen shot 2018-08-08 at 11 44 12

Alternative icon could be a question mark:

screen shot 2018-08-08 at 11 47 36

PR #761, opened 2018-08-09
Sync subject assignees and allow filtering by assignee

### LINKED ISSUE #722: Sync subject assignees
It'd be great to be able to filter and search notifications by issues assigned to you, I think all we need to save is an array of usernames on the subject from the `assignees` field on the [issues api](https://developer.github.com/v3/issues/#get-a-single-issue).

andrew:
Fixes #722, also related to https://github.com/octobox/octobox/pull/752

screen shot 2018-08-09 at 10 13 18

Sidebar menu item defaults to "Assigned to you", where "you" is the current logged in user, but the url param contains the current users login, i.e `octobox.io/?assignee=andrew`, so you can filter by issues/prs assigned to other users by manually change the url and in future the [advanced search](https://github.com/octobox/octobox/issues/742).

## TODO

- [x] Figure out if/how to also support mysql
- [x] Add a way to resync existing subjects to add assignee data


andrew:
Ah, I forgot about mysql, this uses a postgresql only feature 😬

andrew:
If you want to try this out on you're local instance, you'll currently need to delete and resync all subjects, run these two commands in the `$ rails console`:

```ruby

Subject.delete_all
Notification.subjectable.without_subject.find_each{|n| n.send(:update_subject) }
```

BenJam:
LGTM (on postgres)

andrew:
Now grouped with unlabelled and bot filters:

screen shot 2018-08-13 at 09 44 22

andrew:
Also just confirmed that https://github.com/octobox/octobox/pull/756#issuecomment-412444522 happens here too, the only time you'll automatically pull in assignee changes is if you got assigned to it by someone else (although unassigning you by someone else doesn't trigger the same thing)

andrew:
@jules2689 any ideas on how to make this mysql friendly? Storing the array with a serializer would be fine but I'm not sure of a way we can do a sql query to find all the subjects with a certain assignee in the array.

jules2689:
json type? https://dev.mysql.com/doc/refman/5.7/en/group-by-functions.html#function_json-arrayagg

andrew:
JSON type for mysql was ok but was really painful to work with in PostgreSQL, instead opted to just serialise the logins into a string column manually.

jules2689:
what's the potential downsides here? perf?

andrew:
With an index there shouldn’t be a performance hit, biggest downside is the non-standard rails code but that can be hidden away in a serializer

PR #769, opened 2018-08-10
Dark mode

### LINKED ISSUE #699: Toggle between light and dark theme from the settings
 - [x] octobox.io
 - [x] My instance

🐞 **Problem**
I really like Octobox, and it's awesome that it's possible to self-host it. The only missing feature for me is the possibility to switch between a dark/light theme in the settings.

🎯 **Goal**
Dark themes are not just something for edgy people who like the dark, it's actually a good feature to have to avoid straining on the eyes, especially for people that often work in a low-light environment. (For example, I usually start working around 6am while it's still dark, so I try to avoid black text on white background as much as possible).

Something to keep in mind though is the contrast between the background and text colours, as well as font size and other accessibility factors. They should comply with W3C’s AA-level standards.

Since Octobox is a professional tool that is probably mostly used throughout the day for extended periods of time, reducing screen glare is important. That's why Photoshop, Adobe products and even more recently MacOS are introducing dark themes.

💡 **Possible solutions**
Someone actually already made a [dark theme UserStyle](https://userstyles.org/styles/156469/octobox-io-dark-tst2005) for Octobox, which requires the use of the Stylish or Stylus browser extensions, and since it isn't directly within the repository, changes to the UI will break this dark theme implementation.

I think the best way to add this feature would be to have a toggle in the settings, and from the quick look I gave to the data models in the db it looks like we could store it into a boolean value in the `users` table.

As for the sass, I guess we could use variables for the backgrounds, font colors, etc. and compile it once while importing a `light_theme.scss` file that would set those variables for the light version, and once while importing a `dark_theme.scss` file. This way, we generate both sets of css.

Now as for switching to the right css depending on the theme, I'm afraid I don't know enough of how the frontend works for Octobox, but I could look into it.

screenshot 2018-07-23 at 06 25 06

Dark theme implementation from [tst2005](https://github.com/tst2005)

Not started yet. I'd like to get some feedback on the feature request first, as well as maybe some technical insights as to whether or not the solutions I described above make sense.

andrew:
First pass at #699, enabling dark mode via a "Theme" option from user settings:

screen shot 2018-08-10 at 09 19 08

Currently using the css from https://userstyles.org/styles/156469/octobox-io-dark-tst2005 but it inverts everything, including org/user avatars in the sidebar which isn't ideal, would prefer to change just the bits we need to instead but this was an easy first attempt, screenshot below:

screen shot 2018-08-10 at 08 59 46

Also opens up the possibility of adding more themes in the future as well.

Fixes #699


BenJam:
technically this is fine but aesthetically this makes me wince. I'll take a stab at the css later, probably based on something like https://bootswatch.com/darkly/

frenck:
I would love a dark mode ❤️

andrew:
@BenJam got a screenshot of your work in progress?

BenJam:
This is in a workable state now:

screen shot 2018-09-07 at 15 59 20

BenJam:
left an optional, blue tint in there: 

image

andrew:
Related issue that will mean we can darken the checkboxes: https://github.com/octobox/octobox/issues/792

app/helpers/application_helper.rb
```diff
@@ -45,4 +45,9 @@ def used_by_orgs
        elastic src-d alphagov vuejs rails algolia
        shopify WordPress golang)
   end
+
+ 	def current_theme
```
andrew: Indentation's a bit off here

app/views/layouts/application.html.erb
```diff
@@ -26,7 +26,7 @@
       <script>window.bugsnagClient = bugsnag({ apiKey: '<%= Rails.application.secrets.bugsnag_js_api_key %>', releaseStage: '<%= ENV["RAILS_ENV"] %>' })</script>
     <% end %>
 
-    <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
+    <%= stylesheet_link_tag  current_theme ? current_theme : 'light' , media: 'all', 'data-turbolinks-track': 'reload' %>
```
andrew: Can we move this logic into the current_theme helper method?
```ruby
def current_theme 
  current_user.try(:theme) || 'light'
end
```

app/views/notifications/index.html.erb
```diff
@@ -14,8 +14,8 @@
         <div class="card-header sticky-top">
           <% if @notifications.to_a.any? %>
             <div class="per-page btn-group float-right d-none d-md-block" data-toggle="tooltip" data-placement="left" title="Notifications per page">
-              <button type="button" class="btn btn-sm btn-outline-secondary dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                <%= @per_page %>
+              <button type="button" class="btn btn-sm btn-outline-dark dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                <%= per_page = params[:per_page].presence || 20 %>
```
andrew: Looks like this got reverted in a merge, line 18 should be `<%= @per_page %>`

db/schema.rb
```diff
@@ -104,6 +104,7 @@
     t.string "encrypted_app_token"
     t.string "encrypted_app_token_iv"
     t.string "sync_job_id"
+	t.string "theme", default: "light"
```
andrew: Indentation's a bit off here too

BenJam:
ready to roll 👍

pradyunsg:
Yay!

There's a few minor details like the active notification color is not contrasting enough in the dark theme and, subjectively, the grey looks ugly in the light theme and captures too much attention.

@benjam @andrew would filing an issue providing feedback on the design changes made in this PR (and recent ones) be acceptable/useful?

PR #775, opened 2018-08-12
Advanced search queries

andrew:
Closes #742

screen shot 2018-08-12 at 22 23 37

Supported keywords:

- repo (octobox/octobox)
- owner (octobox)
- type (issue, pull request etc)
- reason (mention, author etc)
- label (if `FETCH_SUBJECT` enabled)
- state (open, closed, merged, if `FETCH_SUBJECT` enabled)
- starred (true/false)
- archived (true/false)
- unread (true/false)
- bot:true
- unlabelled:true

Also adds a small visual change in search results to show if a notification has been archived or not:

screen shot 2018-08-12 at 22 18 39

~n.b. related (existing) bug there, can't unarchive things from search results:~

screen shot 2018-08-12 at 22 20 09

Also doesn't highlight the inbox when showing search results (as we also show archived things by default):

screen shot 2018-08-12 at 22 19 00

Multiple, comma separated args aren't supported (`label:bug,design`) atm, hence singular names for all keywords.

No sorting included here either.

One other thing to consider, but probably doesn't need to be solved in this PR is if we turn the keywords into real filters, currently the UI doesn't know that when you search for `owner:octobox` that you're filtering by owner, so the UI looks like this:

screen shot 2018-08-12 at 22 25 00

rather than this:

screen shot 2018-08-12 at 22 25 09

Looking at how gmail does it's search/filtering, the UI does respond to keywords in search, i.e. `is:starred` highlights the starred sidebar item:

screen shot 2018-08-12 at 22 26 02

Also interesting there is that clicking on the `Starred` sidebar adds `is:starred` to the search box, which might be something for us to consider, although you can't click on multiple sidebar items to keep filtering down in gmail.

There's also slightly nasty hack to the `searrrch` gem to add support for `/`, `-` and `.` in filter values, I will attempt to contribute a patch back but it's not been touched 2 years so may not get merged.

TODO:

- [x] Replace searrrch hack
- [x] Docs
- [x] Tests

PR #781, opened 2018-08-13
Sync repository visibility along with subjects

andrew:
For future features, it'll be very useful to know if the repository a notification has come from is open source or a private repo, this pr syncs basic repo details in a similar fashion to subject syncing (also makes it a good candidate for syncing multiple via the graphql api soon).

Only visible UI change is in the repo list on the sidebar:

screen shot 2018-08-13 at 16 36 57

If octobox can't access a repo, it assumes it's private.

Todo:

- [x] Fix tests


andrew:
We can potentially also add a sidebar filter for public/private but I don't think it's that useful yet, will probably just add a filter to #775 instead

screen shot 2018-08-13 at 16 49 25

andrew:
@pradyunsg opened a seperate PR for adding the filters: https://github.com/octobox/octobox/pull/783

PR #783, opened 2018-08-14
Repo visibility filter

andrew:
Follow on from #781, adds public/private filter to sidebar:

44043329-892dc486-9f1a-11e8-97b0-c33ea246ef1a


app/views/notifications/_sidebar.html.erb
```diff
@@ -103,6 +103,19 @@
   <% end %>
   <%= menu_separator unless @reasons.empty? %>
 
+  <% if @visiblity %>
+    <% @visiblity.sort_by{|visiblity, count| visiblity.to_s }.each do |visiblity, count| %>
+      <%= filter_link :is_private, visiblity, count do %>
+        <% if visiblity %>
```
rafaelfranca: Visibility will be always true here given `nil.to_s` is true. I think you want to check if visibility is 'true' here.

rafaelfranca: You can see by the result in the page.

screen shot 2018-08-29 at 14 07 19

andrew: Opened as a separate issue here too: https://github.com/octobox/octobox/issues/847

PR #806, opened 2018-08-17
Install Octobox as a GitHub App

### LINKED ISSUE #472: Support for optionally enabling downloading subjects

andrew:
Working towards #472 

This PR enables Octobox to gain access to issues and pull request data to individual repos rather than needing to request `repo` scope for the user. It works by combining both the existing Oauth App which requests `notifications` scope for the logged in user and allows them to install a new [GitHub App](https://developer.github.com/apps/) on orgs or repositories. 

The user then is requested to login with the GitHub App which stores a second encrypted token on their account (`app_token`) which is used for making requests to sync subjects on repos they have access to that have the app installed.

This builds on https://github.com/octobox/octobox/pull/781 to keep track of which repositories have the app installed on it (via updated automatically from the GitHub app webhook) so we can prompt for notifications that don't have the app to install it to enable subject syncing.

The webhook will also solve a niggling issue where changes to assignees and labels don't get synced automatically as they don't create an update to a notification, see https://github.com/octobox/octobox/pull/756#issuecomment-412444522 and https://github.com/octobox/octobox/pull/761#issue-207251988, as well as actions performed by yourself (opening, closing, commenting) won't trigger a subject sync as there's no notification update. 

The GitHub App comes with an automatic webhook that can tell the install about all changes related to issues and pull requests.

screen shot 2018-08-17 at 18 00 28

All users who have notifications from a repo with the app installed will see the extra status, author, assignee etc details, only one person needs to install the app per repo.

Also the only way to access the new [Checks API](https://developer.github.com/v3/checks/) is to call it from a GitHub App, so we'll be able to add support for that in the future too.

This won't be the default way to run Octobox, it's primarily aimed at enabling subject syncing on https://octobox.io, all existing self-hosted instances will continue to work as usual during the upgrade but you should be able to setup a GitHub app version if you wish.

<hr>

Things still to do:

- [x] Delete subjects when a GitHub App is uninstalled from a repository
- [x] Handling of all relevant web hook types
- [x] Store permissions on AppInstallations
- [x] Handle `github_app_authorization` webhook


andrew:
Adding a note to myself for something to think about for a later pull request.

The `sync_involved_users` method will enable us to automatically and instantly sync users notifications for existing subjects that have been updated but that won't work for brand new subjects.

With the GitHub app webhook, we'll likely be notified of new subjects before we'll have synced any new notification records for that subject. 

A few ways we could guess at users who are likely to get notified of a new subject:

- sync a users [watched repositories](https://developer.github.com/v3/activity/watching/)
- look for @mentions in the body of the subject
- assignees and requested reviewers
- users who've been notified of other subjects in the same repository (inferring the watchers)

andrew:
Another note: Having a sidekiq worker running is likely going to be a requirement to running Octobox as a GitHub app

BenJam:
🙌

andrew:
@cmluciano yeah docs are still on my todo list, I wanted to kick the tyres on this in production sooner rather than later

andrew:
@cmluciano opened a separate issue for the docs so it doesn't get lost: https://github.com/octobox/octobox/issues/850

andrew:
@cmluciano https://github.com/octobox/octobox/pull/851

PR #823, opened 2018-08-20
Show action buttons on hover of notification row

### LINKED ISSUE #648: Show icons to act on a notification when hovering over it
 - ✅ octobox.io
 - My instance (not yet; any way to deploy it for free?)

**Problem**

Show archive and mute buttons for an entry when hovering over it.

**Goal**

This would optimize my workflow where I open an issue, check and see if there's anything actionable. If I'm not interested, I mute it otherwise, once I close an issue, I archive it. I'd guess this to be a common workflow among the users.

Currently the last part of it (archiving/muting a single notification) takes 2 fairly separated, small targets to be clicked. Getting a single icon that can be clicked when hovering over a notification would help simplify that workflow.

In terms of UX, archiving/muting a single notification is likely a common operation and making it easier to do is a good thing. :)

**Possible solutions**

Might be a purely frontend change. If so, it'd just need to have an element that gets shown when hovering over an entry and then that when clicked dispatches the same request as selecting and archiving would; I guess.

andrew:
Fixes #648

Screen shot:

screen shot 2018-08-20 at 17 22 12


pradyunsg:
LGTM. Haven't tested this out yet but the SS seems perfect. :)

pradyunsg:
Would clicking the buttons result in the page being reloaded? If so, would scroll position be retained?

andrew:
@pradyunsg both of those things should definitely be added, good thinking

andrew:
+1 from @bengourley on this feature over lunch today

andrew:
Going to 🚢 this as is now, where hitting one of the inline buttons triggers a page reload, will improve in a latter PR with more javascript

BenJam:
there's always room for more javascript 😬

bengourley:
MOAR!

PR #834, opened 2018-08-23
Add documentation page

BenJam:
* adds a /documentation page
* adds app_install_url env. variable and hides some docs behind it
* moves keyboard mappings and search help to shared partials
* hides the search bar if there are no notifications to present

Needs to fix up a bug with the sticky-top menu bar that's added on mobiles, but I suspect there's a bigger task to fix up some of the weird scrolling behaviour across the app.


app/helpers/application_helper.rb
```diff
@@ -20,6 +20,10 @@ def flash_messages
     nil
   end
 
+  def repo_scope_modal()
```
andrew: No brackets needed here

lib/octobox/configurator.rb
```diff
@@ -118,6 +118,10 @@ def redis_url
       @redis_url = ENV.fetch("REDIS_URL", "redis://localhost:6379")
     end
 
+    def app_install_url
+      @app_install_url || ENV['app_install_url'].present?
```
andrew: `ENV['app_install_url']` should be `ENV['APP_INSTALL_URL']`

PR #709, opened 2018-07-30
Thread view

andrew:
⚠️ Work in progress ⚠️ 

Adding a view to see all the comments on an issue/pull request if subject syncing is enabled, as discussed with @csexton in [gitter](https://gitter.im/octobox/octobox)

screen-shot-2018-07-25-at-11 43 23

Still quite a lot to do here, a few things of note:

- [ ] collapse comments that have already been seen (anything before `last_read_at`)
- [ ] Allow replies directly from that view
- [ ] Hook up the archive, star and mute buttons to actually work
- [x] ensure the visuals work correctly with the bootstrap 4 that's currently in progress https://github.com/octobox/octobox/pull/366
- [ ] Tests!
- [ ] Make subject title clickable

This also allows us to show all the people involved in an issue/pr and show them on the main page (cc @chrisarcand as he mentioned something about that data on twitter the other day):

screen_shot_2018-07-25_at_16 24 30

Syncing speed can also be greatly improved by using a similar approach to https://github.com/octobox/octobox/pull/482

Issue #268, opened 2017-01-19
Javascript tests


WillAbides:
I'd like to add some tests for functions in application.js.  The need became apparent in this [unfortunate](https://github.com/octobox/octobox/pull/248) [series](https://github.com/octobox/octobox/pull/252) of [PRs](https://github.com/octobox/octobox/pull/253).

I'm thinking of using [mocha](https://github.com/mochajs/mocha), but I'm up for other suggestions.

andrew:
👍

johnthepink:
I've been using [Jest](http://facebook.github.io/jest/) recently. It has a great command line interface, it's very fast, sane defaults, and super easy mocking.

andrew:
Needs to work with Ruby on Rails test suite btw for stubbing and mocking http requests

WillAbides:
Since this has some interest, I created a WIP PR to show where I've gotten with this.  #287

technicalpickles:
Would javascript unit tests guard against the type of problems you are trying to prevent? Or do you think it needs full browser tests with something like capybara + selenium/poltergeist?

WillAbides:
The particular issues that caused me to open this issue would be caught by unit tests, but browser tests would catch a lot more.

samjacobclift:
Why not both :D unit and browser ones, we could add unit ones for now and then add browsers ones if need be

PR #844, opened 2018-08-29
Add visual acceptance tests

BenJam:
I'd like to do some refactoring of the frontend scrolling/transitions, so I started at looking at Percy to do visual diffs. I also suspect that #709 will be refactored as a new panel rather than page, which means a lot more SPA-like activity and clicky-clicky testing 😅

Unfortunately the default `rack_test` driver for Capybara doesn't support window resizing which we need to simulate the sidebar (and future thread view) panel(s) so I had to use `:selinium` and `geckodriver` which adds overhead on installation and slows testing. 

I'm not sure what the best approach here should be but it feels like this might encumber those running Octobox locally... Thoughts?


andrew:
If you move the visual tests to a separate folder inside the `test` directory, maybe called `test/visuals`, then you can run them on their own with `rake test:visuals` (or any other folder name like `rake test:models`)

Then we can change the default `rake` task to skip the visual tests but still have them run on travis.

andrew:
Related: https://github.com/octobox/octobox/issues/268

BenJam:
> It's alive! It's ALIVE! Muuhuhahha

BenJam:
I think this is ready to merge.  The purpose here isn't to test the functionality (that's in the minitests) but to do visual diffing over on Percy to pick anything up that we might otherwise miss.

PR #878, opened 2018-09-05
Refactor and improve test coverage on configurator class

andrew:
What started out as a little refactor turned into a slightly bigger one as I remembered that we support [Adding a custom initializer](https://github.com/octobox/octobox/blob/master/docs/INSTALLATION.md#adding-a-custom-initializer):

```ruby
# config/initializers/custom.rb

Octobox.config do |c|
  c.personal_access_tokens_enabled = true
end
```

Although we don't currently have any tests for that or documentation about which config options can be overridden by that block.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,7 +25,7 @@ class ApplicationController < ActionController::Base
   private
 
   def display_subject?
-    return true if Octobox.config.fetch_subject
+    return true if Octobox.fetch_subject?
     Octobox.config.github_app && current_user && current_user.app_token.present?
   end
 
--- a/lib/octobox/configurator.rb
+++ b/lib/octobox/configurator.rb
@@ -44,7 +44,7 @@ module Octobox
     attr_writer :github_app
 
     def fetch_subject
-      @fetch_subject || (ENV['FETCH_SUBJECT'].try(:downcase) == "true")
+      @fetch_subject || env_boolean('FETCH_SUBJECT')
     end
     attr_writer :fetch_subject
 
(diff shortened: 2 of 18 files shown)


app/controllers/application_controller.rb
```diff
@@ -25,7 +25,7 @@ class ApplicationController < ActionController::Base
   private
 
   def display_subject?
-    return true if Octobox.config.fetch_subject
+    return true if Octobox.fetch_subject?
```
andrew: Moved a number of boolean checks onto the `Octobox` class with a question mark which reference the non-questionmark config options

lib/octobox/configurator.rb
```diff
@@ -44,7 +44,7 @@ def github_app
     attr_writer :github_app
 
     def fetch_subject
-      @fetch_subject || (ENV['FETCH_SUBJECT'].try(:downcase) == "true")
+      @fetch_subject || env_boolean('FETCH_SUBJECT')
```
andrew: `env_boolean` helper that allows switching options on or off by passing `true/false` rather than toggling via value/blank

lib/octobox/configurator.rb
```diff
@@ -53,44 +53,32 @@ def subjects_enabled?
...
 
     def personal_access_tokens_enabled
-      @personal_access_tokens_enabled || ENV['PERSONAL_ACCESS_TOKENS_ENABLED'].present?
+      @personal_access_tokens_enabled || env_boolean('PERSONAL_ACCESS_TOKENS_ENABLED')
     end
     attr_writer :personal_access_tokens_enabled
 
     def minimum_refresh_interval
-      @minimum_refresh_interval || ENV['MINIMUM_REFRESH_INTERVAL'].to_i
+      @minimum_refresh_interval || env_integer('MINIMUM_REFRESH_INTERVAL')
```
andrew: `env_integer` helper for integer ENV vars that handles env being nil as well

lib/octobox/configurator.rb
```diff
@@ -53,44 +53,32 @@ def subjects_enabled?
...
 
     def max_notifications_to_sync
-      if @max_notifications_to_sync
-        @max_notifications_to_sync
-      elsif ENV['MAX_NOTIFICATIONS_TO_SYNC'].present?
-        ENV['MAX_NOTIFICATIONS_TO_SYNC'].to_i
-      else
-        500
-      end
+      @max_notifications_to_sync || env_integer('MAX_NOTIFICATIONS_TO_SYNC', 500)
```
andrew: `env_integer` also takes a default value to fall back to

lib/octobox/configurator.rb
```diff
@@ -100,26 +88,23 @@ def sidekiq_schedule_path
...
-      id = @github_team_id || ENV['GITHUB_TEAM_ID']
-      return id.to_i if id.present?
+      @github_team_id || env_integer('GITHUB_TEAM_ID')
     end
     attr_writer :github_team_id
 
     def native_link
       ENV['OCTOBOX_NATIVE_LINK'] || nil
     end
-    attr_writer :native_link
```
andrew: Removed `attr_writer` for methods that won't currently respect those options if overridden, may end up adding them back in and enabling them in the config initializer.

lib/octobox/configurator.rb
```diff
@@ -163,5 +145,15 @@ def github_admin_ids
       return @admin_github_ids = [] unless admin_github_ids.present?
       @github_admin_ids = admin_github_ids.split(',')
     end
+
+    private
+
+    def env_boolean(env_var_name)
+      ENV[env_var_name].try(:downcase) == 'true'
```
jules2689: it would be nice if this was:
```ruby
%w(true 1).include?(ENV[env_var_name].try(:downcase))
```
Because we had `ENV['env'].present?`, it's possible people used other values. 1 or true seem reasonable to me?

PR #880, opened 2018-09-05
Persist per page in cookie

### LINKED ISSUE #779: Sticky per-page settings
I noticed that power users who get loads of daily notifications often set their per-page to 100 🏋️‍♀️ 

Only problem with that is that it's only persisted in the url, so if you go to the settings page or just click the logo you loose your per-page preference. 

We could easily save their per-page preference to their `User` record in the database after they change it from the dropdown and then use as their default for ever more (until they change it).

paulleader:
Stores the `per_page` value set with a param into a cookie of the same name.

Also fixes the template to use the per_page value calclated by the controller rather than just taking the param.

Fixes #779

### DIFF
--- a/app/views/notifications/index.html.erb
+++ b/app/views/notifications/index.html.erb
@@ -15,7 +15,7 @@
           <% if @notifications.to_a.any? %>
             <div class="per-page btn-group float-right d-none d-md-block" data-toggle="tooltip" data-placement="left" title="Notifications per page">
               <button type="button" class="btn btn-sm btn-outline-secondary dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                <%= per_page = params[:per_page].presence || 20 %>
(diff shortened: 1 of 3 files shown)


app/views/notifications/index.html.erb
```diff
@@ -15,7 +15,7 @@
           <% if @notifications.to_a.any? %>
             <div class="per-page btn-group float-right d-none d-md-block" data-toggle="tooltip" data-placement="left" title="Notifications per page">
               <button type="button" class="btn btn-sm btn-outline-secondary dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                <%= per_page = params[:per_page].presence || 20 %>
+                <%= per_page = @per_page || 20 %>
```
andrew: It looks to me like we can just delete this line altogether and replace `per_page.to_i` on line 23 with `@per_page`

PR #891, opened 2018-09-07
Add sorting to notifications

### LINKED ISSUE #813: Sorting options
Pulled out from the [triage view thread](https://github.com/octobox/octobox/pull/752#issuecomment-413916688).

A number of people have mentioned wanting the ability to sort lists in ways other than `updated_at DESC`, accessible via a drop down and search prefix (`sort:last_read_at order:asc`).

List of potential sorting columns:

- created_at
- updated_at (ascending)
- last_read_at
- subject_title

Some more options once these features are complete:

- Comment count https://github.com/octobox/octobox/pull/709
- due date https://github.com/octobox/octobox/issues/812
- muted_at https://github.com/octobox/octobox/issues/811
- "last time I commented" @pradyunsg in https://github.com/octobox/octobox/pull/752#issuecomment-411306052 (semi-related to https://github.com/octobox/octobox/issues/739)

paulleader:
Adds sorting to the notification list for:

- `subject`
- `updated`
- `read`

I've skipped `created_at` for the moment as it doesn't seem to result in a sort order that makes sense. The `updated` time is the same as the one displayed with each notification so it makes sense, but `created_at` just looks messed up.

Resolves #813


paulleader:
@andrew Think this is now working as intended.

andrew:
I think the behavior we want for `created_at` is the subject `created_at`, i.e. when the issue/pr was first opened, so that you can find the newest/oldest, similar to the default sort of GitHub issues:

screen shot 2018-09-09 at 09 08 31

I suspect `notifications.created_at` is a little borked because we currently don't save the remote created_at field, looking at https://github.com/octobox/octobox/blob/master/app/services/download_service.rb#L8-L22, adding that there might just magically fix that.

andrew:
> I suspect notifications.created_at is a little borked because we currently don't save the remote created_at field, looking at https://github.com/octobox/octobox/blob/master/app/services/download_service.rb#L8-L22, adding that there might just magically fix that.

Or not, looks like `created_at` isn't a field in the notifications API: https://github.com/octobox/octobox/blob/master/test/fixtures/files/notifications.json#L41

app/models/search.rb
```diff
@@ -24,11 +24,50 @@ def results
...
   private
 
+  def apply_sort(scope)
+    case sort_by
+    when 'subject'
+      scope.reorder("upper(notifications.subject_title) #{sort_order}")
+    when 'updated'
+      scope.reorder(updated_at: sort_order)
+    when 'read'
+      scope.reorder(last_read_at: sort_order)
```
andrew: @paulleader `last_read_at` is often nil, not sure if that impacts this or not?

paulleader: @andrew Hrrrm, good point, on Postgres it sorts nil to be max, so in descending order you get all the unread ones first.

paulleader: Yeah, mysql treats null as minimum value :/

Will have a think about that and put in something tomorrow evening to make them consistent.

What would make more sense? The most sensible order for last read is probably descending (so most recent first) with unread ones at the end I would think.

Or, we could simply exclude unread ones from that search.

andrew:
Re-enabling is here: https://github.com/octobox/octobox/pull/905

PR #901, opened 2018-09-10
Add Octobox.github_app_client for making API requests as the GitHub App

andrew:
https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#authenticating-as-a-github-app

Useful for getting information on installations without an authenticated user, i.e. in background jobs


andrew:
Just investigating to see if we can use the existing `encryptor` gem dependency instead of adding another openssl wrapper gem.

PR #905, opened 2018-09-11
Bring back #891

andrew:
A quick merge+deploy of #891 broke master and https://octobox.io 

The issue was really a workflow one rather than a code one.

I merged #891 without rebasing it against master, so the tests appeared to be passing but only showed up broken once merged into master.

I'm going to enable some more branch protections on this repo to stop this happening in future.


andrew:
Up-to-date and green status on CI now required for merging to master:

screen shot 2018-09-11 at 14 15 40

PR #908, opened 2018-09-11
Add vulnerability disclosure policy

### LINKED ISSUE #906: Add vulnerability disclosure policy
We should add a policy that makes it clear we support (and authorise) security research on Octobox. Looks like 18F have a sensible, public domain licensed document at https://github.com/18F/vulnerability-disclosure-policy/blob/master/vulnerability-disclosure-policy.md

BenJam:
closes #906


BenJam:
Can do. Should we add an incident and disclosure log at this point too?

andrew:
My preference would be to do that in another PR

andrew:
Speedy 🚢

Issue #919, opened 2018-09-12
Allow filtering by multiple repositories with search prefix


andrew:
Feature requested by @gbp.

For a search of `repo:24pullrequests/24pullrequests,octobox/octobox`, should return results for both repositories. Currently ignores any more after the first repository.

Two bits of code need changing for this, the easy one, remove `.first` from https://github.com/octobox/octobox/blob/master/app/models/search.rb#L72

and the second, harder one, on https://github.com/octobox/octobox/blob/master/app/models/notification.rb#L33 we need to convert an array of strings (of any length) like `["24pullrequests/24pullrequests", "octobox/octobox"]` into an arel query like:

```ruby
Notification.arel_table[:repository_full_name].matches("24pullrequests/24pullrequests"').or(Notification.arel_table[:repository_full_name].matches("octobox/octobox"))
```

Chaining extra `.or(Notification.arel_table[:repository_full_name].matches(repo_full_name)` on the end for every extra repo name in the argument.

n.b. we're using `arel_table[:field].matches` because it switches between LIKE or ILIKE depending on if postgres or mysql is being used.

prashuchaudhary:
@andrew I have made the recommended changes. Please have a look

andrew:
Fixed by #921, thanks @prashuchaudhary!

PR #921, opened 2018-09-12
Addresses #919 Allow filtering by multiple repositories with search p…

prashuchaudhary:
…refix


gbp:
This is great, I've tested this and it works!

One thing I noticed, if you check for `sub_scope == :repo` [here](https://github.com/octobox/octobox/blob/master/app/controllers/notifications_controller.rb#L254) it would also allow this to be used when accessing `/?repo=a%2Fb%2Cb%2Fc`.

prashuchaudhary:
@gbp Making a request like this `/?repo=a%2Fb%2Cb%2Fc` will return No Results, which is correct behavior given that it lists all the Unread notification of a repo. 

What would you suggest should be done here ??

Issue #811, opened 2018-08-18
Store muted_at and allow search filtering by muted:true/false


andrew:
Pulled out from the [triage view thread](https://github.com/octobox/octobox/pull/752#issuecomment-413916688)

We currently don't know/store if a user has previously muted a notification thread, if we store `muted_at` on the notifications table that will allow us to see if and when a notification was muted, and also if there has been activity on it since it was muted (I have a feeling that if you mute a thread and then get mentioned in a thread GitHub still notifies you)

We can then add another prefix to the advanced search `muted:false` (or true) to allow you to filter out issues/prs that you were active on that you're no longer interested in.

prashuchaudhary:
@andrew I have made the recommended changes. Please have a look #923

PR #923, opened 2018-09-13
Addresses #811 Store muted_at and allow search filtering by muted:tru…

prashuchaudhary:
…e/false


db/migrate/20180913111742_add_muted_at_to_notifications.rb
```diff
@@ -0,0 +1,6 @@
+class AddMutedAtToNotifications < ActiveRecord::Migration[5.2]
+  def change
+    add_column :notifications, :muted_at, :datetime, null: true
+    add_index :notifications, :muted_at
+  end
+end
```
andrew: Can you commit the change to `db/schema.rb` that this migration generates as well

app/models/search.rb
```diff
@@ -24,6 +24,7 @@ def results
     res = res.bot_author unless bot_author.nil?
     res = res.unlabelled unless unlabelled.nil?
     res = res.is_private(is_private) unless is_private.nil?
+    res = muted ? res.muted : res.unmuted
```
andrew: Can we change this to act the same as `is_private` line above, so it only filters by muted if the muted: prefix is passed through

app/models/search.rb
```diff
@@ -129,4 +130,9 @@ def is_private
     return nil unless parsed_query[:private].present?
     parsed_query[:private].first.downcase == "true"
   end
+
+  def muted
+    return false unless parsed_query[:muted].present?
+    parsed_query[:muted].first.try(:downcase) == "true"
+  end
```
andrew: Can you add documentation for this prefix to https://github.com/octobox/octobox/blob/master/app/views/shared/_search_prefixes.html.erb#L60

app/controllers/notifications_controller.rb
```diff
@@ -239,7 +239,7 @@ def selected_notifications
     if params[:id] == ['all']
       current_notifications
     else
-      current_user.notifications.where(id: params[:id])
+      current_user.notifications.unmuted.where(id: params[:id])
```
andrew: Same as above, let's keep the default scope the same as before.

app/controllers/notifications_controller.rb
```diff
@@ -104,7 +104,7 @@ def index
   #   { "count" : 1 }
   #
   def unread_count
-    scope = current_user.notifications
+    scope = current_user.notifications.unmuted
```
andrew: Let's keep the default scope the same as before, when notifications are muted, they also get archived so will disappear from the inbox already. This feature is mostly for the advanced search prefixes.

db/migrate/20180913111742_add_muted_at_to_notifications.rb
prashuchaudhary: sure. will do that.

app/models/search.rb
prashuchaudhary: will do that

app/controllers/notifications_controller.rb
prashuchaudhary: will update

app/controllers/notifications_controller.rb
prashuchaudhary: will update

prashuchaudhary:
@andrew made the requested changes

app/models/search.rb
```diff
@@ -24,6 +24,7 @@ def results
     res = res.bot_author unless bot_author.nil?
     res = res.unlabelled unless unlabelled.nil?
     res = res.is_private(is_private) unless is_private.nil?
+    res = res.muted unless is_muted.nil? || !is_muted
```
andrew: Does this work for `muted:false`, i.e `unmuted`?

prashuchaudhary: yes. for both `true` and `false` case

prashuchaudhary: `muted: false` case is the same as having null value for `muted_at` in DB. So I am only applying scope for `muted:true` case

prashuchaudhary: Added test case for `muted:false`

andrew:
@prashuchaudhary I just tweaked your failing test for the behaviour I was looking for `muted:false` should use the `unmuted` scope rather than no scope, I think this is also the same on https://github.com/octobox/octobox/pull/926

test/controllers/notifications_controller_test.rb
```diff
@@ -542,10 +542,10 @@ class NotificationsControllerTest < ActionDispatch::IntegrationTest
   test 'search results can filter by muted:false' do
     sign_in_as(@user)
     Notification.destroy_all
-    create(:notification, user: @user)
+    create(:notification, user: @user, muted_at: Time.current)
     create(:notification, user: @user)
     get '/?q=muted%3Afalse'
-    assert_equal assigns(:notifications).length, 2
```
prashuchaudhary: @andrew `muted:false` will be treated as no muted option provided at all.  So `notifcations.length` will be 2.

andrew: The behaviour that we’d like is for `muted:false` to not return muted notifications

prashuchaudhary: ok. will update the code on both PR's

prashuchaudhary:
@andrew made changes for unmated scope as discussed

andrew:
Awesome, thanks @prashuchaudhary!

PR #924, opened 2018-09-13
Fix svg fill color issues

BenJam:
... bored of one commit fix PRs yet? Leave this one open and add anything else spotted. Now about that #844 PR ...


andrew:
Another related bug, the X on the active sidebar item has an extra background:

screen shot 2018-09-13 at 15 02 40

PR #926, opened 2018-09-13
Addresses #828 Sync locked details onto subjects

### LINKED ISSUE #828: Sync locked details onto subjects
You can lock conversations on GitHub and provide a reason why: https://help.github.com/articles/locking-conversations/

We should add a `locked` field to the `subjects` table and update that field when syncing subjects.

Then we can add a search prefix for filtering `locked:true` or `locked:false`

prashuchaudhary:
Fixes #828

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -24,12 +24,17 @@ class Search
     res = res.bot_author unless bot_author.nil?
     res = res.unlabelled unless unlabelled.nil?
     res = res.is_private(is_private) unless is_private.nil?
-    res = res.locked unless is_locked.nil? || !is_locked
+    res = lock_conditionally(res)
     res = apply_sort(res)
     res
   end
 
   private
+
+  def lock_conditionally(scope)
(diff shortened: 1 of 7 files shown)


prashuchaudhary:
@andrew please review the changes

app/models/search.rb
```diff
@@ -24,6 +24,7 @@ def results
     res = res.bot_author unless bot_author.nil?
     res = res.unlabelled unless unlabelled.nil?
     res = res.is_private(is_private) unless is_private.nil?
+    res = res.locked unless is_locked.nil? || !is_locked
```
andrew: Does this work for `locked:false`?

prashuchaudhary: yes. it works for that too

prashuchaudhary:
@andrew will add test cases

prashuchaudhary:
Added unit test case for Locked flag.

andrew:
Thanks again @prashuchaudhary 👏

PR #929, opened 2018-09-14
Add validation of unique urls in subjects table

### LINKED ISSUE #928: Same notification appears multiple times in the list
 - [x] octobox.io

🐞 **Problem**
There is one notification from a private repository that appears repeated 6 times in my list:
2018-09-14-110326_617x511_scrot

If you need more details, I can provide the exact repository and issue number in private (e.g. on gitter).

andrew:
Fixes #928, which was caused by multiple subjects with the same `url` existing in the database.

Doesn't add unique constraint to the database index because it will fail for any installation that has a duplicate in the subjects table.

Also adds a rake task to clear up duplicate subjects.


pradyunsg:
Should the unique constraint be added eventually?

andrew:
Ideally, but right now anyone who would try to run a migration to add the unique index would get an error if they had just one duplicate, and would need to run the `deduplicate_subjects` rake task.

PR #943, opened 2018-09-15
feature(docker): Adapt Docker image to run sidekiq when enabled

rhuss:
Modified Docker image to start check OCTOBOX_SIDEKIQ_SCHEDULE_ENABLED
and start foreman with rails+sidekiq if set. If not set, then
rails is start as usual.

This is alternative approach to #942 and implements a unified approach
for the Docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,4 +25,7 @@ RUN RAILS_ENV=development bin/rails api_docs:generate
 RUN chgrp -R 0 /usr/src/app \
  && chmod -R g=u /usr/src/app
 
+# Install Foreman
+RUN gem install foreman
+
 CMD ["bin/docker-start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,31 +1,41 @@
 FROM ruby:2.5.1-alpine
-RUN apk add --update \
-  build-base \
-  netcat-openbsd \
-  git \
-  nodejs \
-  postgresql-dev \
-  mysql-dev \
-  tzdata \
-  curl-dev \
-  && rm -rf /var/cache/apk/*
-# throw errors if Gemfile has been modified since Gemfile.lock
-RUN bundle config --global frozen 1
-RUN mkdir -p /usr/src/app
+
+# Install and update all dependencies (os, ruby)
 WORKDIR /usr/src/app
 
+# =============================================
+# System layer
+
+# Will invalidate cache as soon a the Gemfile changes
 COPY Gemfile Gemfile.lock /usr/src/app/
-RUN bundle install --without test production --jobs 2
 
+# * Setup system
+# * Install Ruby dependencies
+RUN apk add --update \
+    build-base \
+    netcat-openbsd \
+    git \
+    nodejs \
+    postgresql-dev \
+    mysql-dev \
+    tzdata \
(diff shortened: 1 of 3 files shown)


chrisarcand:
Oh hey you already implemented it :D Hah, yeah I like this one better.

ℹ️ I'm working around this area too, cleaning up all the docker stuffs.

Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
 COPY Gemfile Gemfile.lock /usr/src/app/
-RUN bundle install --without test production --jobs 2
```
chrisarcand: 👍

rhuss:
Sorry if there is some overlap, however, I needed it for my OpenShift Online installation as I'm running in some timeout theses days when refreshing (there is a hard 30s timeout on the OpenShift route, which leads to 502 errors quite often).

Regardless of which solution is chosen, really would love a single Docker image with background fetching enabled.

Dockerfile
andrew: Still need to bundle install something, maybe just remove `production` from that line

bin/docker-start
```diff
@@ -10,4 +10,8 @@ done
 bundle exec rake db:migrate
 rm -rf tmp/pids
 
-exec rails s -b 0.0.0.0
+if [ "${OCTOBOX_SIDEKIQ_SCHEDULE_ENABLED}" == "true" ]; then
```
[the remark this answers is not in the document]
rhuss: Good point. I use now the definition of 'boolness' which I found here https://github.com/octobox/octobox/blob/b1459701d753a5ececc922c3a05bf6824585d961/lib/octobox/configurator.rb#L161-L163 i.e. true for "true" and "1", false otherwise.

chrisarcand:
Looks great! I'll base my upcoming changes off of what you've done here.

andrew:
👍

chrisarcand:
Hmmm @rhuss I've merged this into my stuff and it looks like the sidekiq worker process isn't getting `REDIS_URL` set properly (I enabled background jobs). Is this something to do with it now being managed in a subprocess by Foreman?

rhuss:
for me it works when setting the env from the outside but I have to set the scheme ro `redis://`. In my case, its 'redis://localhost'. 

'localhost' alone was not good enough.

chrisarcand:
Are you using Docker Compose? I'm attempting to let it use the internal network here: https://github.com/octobox/octobox/blob/master/docker-compose.yml#L27

chrisarcand:
And are you building from your own image or taking the one from Docker Hub?

Yeah, the more I think on this the more I realize Sidekiq should probably be in it's own container; the issue is that using Foreman means configuring REDIS_URL for Sidekiq via the environment involves using `.env`; this is bad because you have to have that within the image, so you're now forced to build your own.

I might be missing something, so correct me if I'm wrong, but I can't figure out how to run this in production with Sidekiq *without* building my own image from source.

chrisarcand:
That was it exactly! Thanks @rhuss for the detailed explanation. I guess it should have been that way from the start, I'll update it.

PR #947, opened 2018-09-16
Add inbox:true search prefix

andrew:
A handy alias for `archived:false`


andrew:
cc @paulleader as we talked about this on Thursday evening.

PR #951, opened 2018-09-16
JS reorganization 

andrew:
Moving all the Octobox specific js functions into an `Octobox` module in `octobox.js` which are then used to initialize the frontend in `application.js`


andrew:
Almost ready to merge, needs some more manual testing as we don't have https://github.com/octobox/octobox/issues/268 yet

PR #964, opened 2018-09-18
Marketplace support

andrew:
Adds support for making the Octobox.io Github app available on the [GitHub Marketplace](https://github.com/marketplace/), we're going through the approval process at the moment. 

From a functionality point of view on Octobox.io, this limits the display of extra subject data (i.e. state, author, assignee and labels) on notifications from private repositories unless they have a subscription setup in the marketplace for a plan with a name that contains the word `private`.

It shouldn't affect any other installations, other than adding a couple of empty database tables.


BenJam:
🥇

PR #968, opened 2018-09-19
archived shouldn't be nil

andrew:
⚠️ work in progress ⚠️ 

I noticed a few notifications on https://octobox.io have got in a weird state where `archived: nil`, which means they don't get shown in the Inbox or Archived views (because NULL is not false-y in sql)

Ideally this will be fixed with a not null restriction on the database column but for now we'll handle existing bad data better and clean it up when syncing.


andrew:
This hasn't worked 100%, still not found the source of the nil yet either but they are still showing up on https://octobox.io

PR #970, opened 2018-09-19
Remove async param from sync button

andrew:
We only ever passed the `async` param when `Octobox.background_jobs_enabled?` was true, so it was kind of redundant. 

There's more than can be refactored in the `sync` controller action but that broke a number of test around error responses from back before we had sidekiq syncing so I'll save that for another day.


chrisarcand:
Good call.

PR #972, opened 2018-09-19
Experiment with caching rvm directory on travis

andrew:
Possibly shaves ~16 seconds off the build time if it works.


andrew:
Doesn't appear to work 😢

PR #976, opened 2018-09-20
fixes #973 Better handling of renamed/transferred repositories #973

### LINKED ISSUE #973: Better handling of renamed/transferred repositories
If a repository is renamed or moved from one owner to another we currently don't do a good job of updating notifications `repository_full_name` or `owner_name` fields (unless the notifications are still in the user's github inbox, which is unlikely for users that have been on the site for over a month).

We do store the repository's remote id against each notification as well so we can potentially use that to find all notifications that need updating.

Probably the best place to put code to handle this is in an `after_save` callback on the Repository model that checks to see if the `full_name` has changed and then finds all notifications with `where(repository_id: repository.github_id)`  and update both `repository_full_name` and `owner_name` fields.

prashuchaudhary:
* Create a module to handle Notifications update if any of the repository is rename/transferred
* Added `activerecord-import` gem for doing bulk update
* Bulk update is done in batch of 1000
* Added Test case


app/models/repository.rb
```diff
@@ -1,4 +1,7 @@
 class Repository < ApplicationRecord
+
+	include Octobox::Repository::UpdateNotificationRepositoryName
```
andrew: Indentation is a little off here

lib/octobox/repository/update_notification_repository_name.rb
```diff
@@ -0,0 +1,31 @@
+module Octobox
```
andrew: Indentation is a little off here too

andrew: Should be 2 spaces

andrew:
Looks like the tests are broken for mysql

lib/octobox/repository/update_notification_repository_name.rb
```diff
@@ -0,0 +1,31 @@
...
+					notification.repository_owner_name =  self.owner
+				end
+
+        # bulk update using activerecord-import gem
+        # on duplicate key option will update the notification attrs instead to creating one
+				Notification.import updated_notifications, :batch_size => 1000,
+				on_duplicate_key_update: {
+					conflict_target: [:id], columns: [:repository_full_name, :repository_owner_name]
+				}
+			end
```
andrew: Would using `notifications.update_all(repository_full_name: self.full_name, repository_owner_name: self.owner)` be possible here instead of an import?

prashuchaudhary:
Will update the test

prashuchaudhary:
@andrew I checked that but This method constructs a single SQL UPDATE
statement and sends it straight to the database. It does not instantiate
the involved models and it does not trigger Active Record
<https://apidock.com/rails/ActiveRecord/Base> callbacks or validations.

So it will not update the updated_at timestamps

andrew:
In this case not updating the `updated_at` timestamp is what we want, as the `updated_at` field mirrors the data from GitHub rather than when we last changed/saved that record locally.

prashuchaudhary:
No prob. Will update using thay method and remove the activerecord-import
one

PR #997, opened 2018-09-23
Don't fire keyboard shortcuts when ctrl/cmd/alt/shift held down

andrew:
Stops octobox from starting a sync when the page is refresh with cmd+r


pradyunsg:
This is awesome!

PR #1000, opened 2018-09-23
fix Alignment issues with Octobox Selected Option for Sidebar

prashuchaudhary:
* Alignment of `X` octokit icon was not proper so fixed that. Checked on Chrome, Firefox, Safari


prashuchaudhary:
@andrew cosmetic fix. please review

andrew:
Good spot!

PR #1007, opened 2018-09-24
Remove username from header

andrew:
screen shot 2018-09-24 at 12 46 39


andrew:
Also added a title for the unusual case where someone has two github accounts with the same avatar:

screen shot 2018-09-24 at 12 56 33

Issue #818, opened 2018-08-20
Labels belong to repositories


andrew:
We currently store a complete set of labels for each subject, which actually ends up duplicating a lot of labels.

In reality, labels belong to a repository and are then many-to-many to subjects, and since https://github.com/octobox/octobox/pull/781 has been merged we can actually connect those dots properly.

That will mean we can update the colour/name for a label from one subject sync and it automatically be updated for all others sharing the same label on that repo.

Semi-related to https://github.com/octobox/octobox/issues/808 as label syncing isn't working right any the moment anyway.

PR #1009, opened 2018-09-25
Labels belong to repositories fixes ISSUE 818

prashuchaudhary:
# Part 1 - Task Breakdown

- [x] Create a new join table `SubjectLabel` for mapping Subject to Labels. 
- [x] Add Repository to Label table; associate is to Repository Model.
- [x] Start Inserting Data in SubjectLabel when we are association Label to Subject. Also, update repository mapping during this.
- [x] Create a rake task to `populate SubjectLabel table with the OLD DATA` and `Update mapping of Labels to Repository`. 
- [x] Move OLD code for associate Label to Subject into a separate module so that if we have to revert back if we only have to make some minor changes.
- [x] Fix existing test cases
- [x] Check Queries on Subject and Labels and Update if necessary
- [x] Write new test cases

 ### RAKE TASK 
 Rake task will find unique Label based on `GITHUB_ID` ordered by the `created_at ASC` and create a mapping between `Subject and Labels`. Also update the Label Repository Mapping.

# Part 2 - Will be tables cleanup.
- [ ] Cleanup OLD labels and Dropping the Subject Column will be done afterwards in case we have to revert back.


andrew:
This is looking good, the slightly tricky part is going to be handling part 2 for users like @jules2689, @chrisarcand and @tarebyte who are going to be upgrading at a different rate to what I'll be doing on https://octobox.io

It might make sense to put the raw sql rake task as part of the migration so there's a nice atomic upgrade without depending on activerecord relation logic.

andrew:
A related performance improvement I was thinking was to cache/denormalize the label name strings onto subjects so when filtering by a label string we can avoid joining across either the labels table (and after this, also joining across the repositories table)

prashuchaudhary:
will move the code and create a migration for Updating the Subject Labels

prashuchaudhary:
@andrew can you please review the changes?

 I have added test cases for all use cases and checked for existing behavior.

andrew:
@prashuchaudhary looks like the tests are broken on mysql

prashuchaudhary:
@andrew checking the tests.

prashuchaudhary:
fixed the broken tests.

prashuchaudhary:
fixes #818

prashuchaudhary:
@andrew can you review these changes ?

prashuchaudhary:
regarding the suggestion to denormalize the `labels.name` to subjects table, we will always
 be doing `like queries`. so it will also be relatively slow ??

BenJam:
Same here @prashuchaudhary.

Issue #778, opened 2018-08-13
Do Not Disturb hours


andrew:
Converting [this tweet](https://twitter.com/teabass/status/1026726387103793152) into an issue:

![screen shot 2018-08-13 at 11 27 05](https://user-images.githubusercontent.com/1060/44026701-d4267138-9eeb-11e8-9358-0ff410a12c98.png)

In particular, being able to hide or ignore "work" repos/orgs outside of work hours. For example, in evenings/weekends I want to see notifications from my side projects without having to look at work-related items.

gnikonorov:
Hi @andrew, could I implement this feature?

gnikonorov:
Thanks @andrew! I'll get this going now.

gnikonorov:
Hi @andrew, I was able to set up `octobox` locally with the instructions given in the readme, and pointed it to http://162.216.19.220:3000/.

Unfortunately it looks like only one repository is showing up. I know this is a vague question, but did I do something wrong? As far as I can tell, I followed all the instructions properly, and have all notification options turned on in GitHub.

PragTob:
:wave: 

As I'm just starting to give octobox another shot I'm not very familiar with the existing UI concepts - that might be good or bad ;)

Coming from this tweet:

![selection_088](https://user-images.githubusercontent.com/606517/45836836-28699500-bd0e-11e8-8797-cb5d8815f679.png)

Generally speaking I personally usually want to look at either my work repos or my OSS repos. Just a button/tab/thing to toggle them would be all I need and ever wanted. It's fine to do it by ORG. So right now for me just work: github.com/liefery and OSS everything else would be a huge win for me.

Time doesn't play a role for me - I might check OSS during my lunch break. Or I might work late as I did a huge break during the day or am making up for time (my work times are very flexible).

I could see the feature creep/usefulness of being more flexible like:
- "also show these OSS repos in work context" (ruby advisory db would be an example for me)
- "only show these projects in my OSS inbox" (projects of which I consider myself a maintainer and hence are more important to me) - this would probably be the most useful to me personally among all the ones I mention here
[...]

PR #1011, opened 2018-09-25
Add user and org as aliases for owner search prefix

andrew:
Re https://github.com/octobox/octobox/issues/778#issuecomment-424093900

I've repeatedly mistaken done searches for `org:octobox` and `user:andrew` when the actual prefix is `owner`. I'm not going to document them but just make them work as expected for people who make the same mistake.

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -98,11 +98,11 @@ class Search
   end
 
   def owner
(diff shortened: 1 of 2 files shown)


andrew:
@pradyunsg having three documented names for the same thing feels too confusing to me.

PR #1014, opened 2018-09-25
Small behaviour change to seleting a search history item

andrew:
Always replace the current search rather than appending it, also fixed a little bit of invalid html.

Follow on from #960


prashuchaudhary:
👍

Issue #963, opened 2018-09-18
Refactor and unify search and filtering options into  single concept


paulleader:
🐞 **Problem**
At the moment, `NotificationsController` deals with raw params, pulling out search and filtering options as needed, and treating the search string (the `q` parameter) as a completely separate concern. Options provided as url parameters are not reflected in the search string. This is preventing us from:

- Converting url options to search string options (e.g. instead of `?reason=author` have the search string `reason:author`)
- Removing the filter bar in favour of the search bar (#952)

🎯 **Goal**
This refactoring aims to combine all search and filtering option handling into a single concept, as a precursor to enabling the above changes.

📋  **Steps (probably one PR per step)**
- [ ] Replace direct access to the various search and filter params with a delegator that wraps the `params` object (`SearchOptions`).
- [ ] Replace `[]` based access to these parameters with methods for each parameter.
- [ ] Add parsing of the query string (`q`) to `SearchOptions`, and expose all the available options as accessors by delegating to `SearchParser`. Unifying access to options that can be specified through either url params or query string options with the string version taking precedence.
- [ ] Update the `Search` class to take a `SearchOptions` object rather than the raw query string.
- [ ] Move search and scoping logic from the `NotificationsController` into the `Search` class, reducing the complexity of the controller.

paulleader:
@andrew Would you mind assigning this to me please :)

andrew:
Excellent write up @paulleader, excited to see this happen and unify the search query with the url query params 🎉

PR #955, opened 2018-09-17
Saved searches

andrew:
⚠️ Work in progress ⚠️ 

TODO:

- proper interface for adding, editing and removing saved searches
- link to saved search settings page
- link/button to save current search from search results page

PR #1015, opened 2018-09-25
Prefill active search query based on filter params

andrew:
Sorry @paulleader, I ended up accidentally half implementing https://github.com/octobox/octobox/issues/963 whilst working on https://github.com/octobox/octobox/pull/955.

The part that I ran into was that I wanted to show `inbox:true` in the search box when you load your inbox with zero url parameters, i.e. https://octobox.io, but that opened a whole can of worms that resulted in this branch.

The Search class has been expanded to accept a `params` hash and then combine the filter options with the parsed query from `SearchParser` into a unified set of filters and buckets that were applied to the page you are currently looking at.

That unified set can then be turned back into a search query string to be inserted into the search text field, as well as introspecting each possible filter/prefix to see if it's present and if so, what value does it have.

The result looks like below regardless of if a filter came from a url parameter `?reason=mention` or a search query prefix `reason:mention`:

screen shot 2018-09-25 at 18 57 39

The extra Search class functionality works pretty well but so far I've only converted half the notifications#index view to use it, the controller action doesn't use it for deciding which parameters to apply yet and the filter links in the sidebar still link to url parameter style filters rather than a modified version of the current page's search query.

Ideally we'll be able to push more Notifications controller logic into the Search class, where it can calculate the requested filters, paginated notifications for a user that match those filters, facets with counts for the sidebar and an interface to add/remove the current filters and get an updated url/query back.

Having all that filter/query logic unified in `Search` would also be helpful if we start to do "suggest as you type" searches, knowing which combinations of prefixes are available to the user as they start to type but that's for another day.


paulleader:
@andrew No probs, have been really struggling to make time these last two weeks.

PR #1030, opened 2018-09-27
CSS cleanup

andrew:
- Remove/combine unused styles and selectors
- Removed a double nested media query
- Delete parts of the dark theme that don't apply
- Don't render empty div when no flash messages to display


BenJam:
lgtm.

PR #1047, opened 2018-10-03
Add repo name as title attribute to filter items

axelerator:
Repository names in our case have a common prefix.
In the sidebar the repo names are `overflow:hidden` and I can only see the beginning of the names which are in a lot cases the same.

By providing the full repo name also as a title attribute we add a non-invasive way to get the info just by hovering with the cursor over the filter item.

I added a optional parameter to the `sidebar_filter_link` so it should not have any impact on existing code.
I would suggest to move to keyword args for helpers like this, since with a lot of optional params it becomes more ugly and error prone to call the method with sparsely filled parameters.

I did not find any tests for the notifications_helper and it was a tiny change so I did not add tests.

### DIFF
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -177,25 +177,25 @@ module NotificationsHelper
   end
 
   def filter_link(param, value, count)
-    sidebar_filter_link(params[param] == value.to_s, param, value, count) do
+    sidebar_filter_link(active: params[param] == value.to_s, param: param, value: value, count: count) do
       yield
     end
   end
 
   def org_filter_link(param, value)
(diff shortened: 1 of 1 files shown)


andrew:
Thanks for this @axelerator 👍 I don't think it would hurt to add the value as a title to every `sidebar_filter_link`. 

That also helps avoid adding another argument to that method which is getting pretty unruly!

axelerator:
Ahh ok I misread first - I tried adding the value as a title but that makes for some pretty weird titles (`true/false` for `read` i.e.)

I added a commit how it would look with keyword args.

andrew:
👌

PR #1068, opened 2018-10-09
Shim in svg support for twbs

BenJam:
Shims in svg fills for theme-colors in twbs:

* text-classes
* btn classes
* btn-outline classes


BenJam:
can haz merge?

andrew:
@BenJam just checked this out, only difference I noticed was that the icons in the user dropdown are darker than before, was that intentional?

screen shot 2018-10-10 at 11 17 15

BenJam:
> was that intentional?

intentional in so much as they match, yup 👍

PR #1076, opened 2018-10-10
Fix some missing colors

BenJam:
TODO build better visual tests 😅


BenJam:
actually merge this, I'll fix sidebar elsewhere.

PR #1079, opened 2018-10-10
Fixed sidebar colors

BenJam:
stripped back some of the unneeded spans
added svg colours for all nav-pills 
make all the text and fill classes that get clobbered by the above `!important`


andrew:
Goodbye `.star-active`, you served us well 👏

Issue #1036, opened 2018-10-01
Sync button stays black (focus state) after sync has finished


BenJam:
![image](https://user-images.githubusercontent.com/158833/46282744-3d45f400-c56a-11e8-8869-a709ff56c779.png)

I've taken a look at this before. I'm not sure whether the `e.preventDefault()` call should stop the focus state. I have tried setting `.blur()` after sync is fired to no avail. The only reason the state is getting reset is when the page reloads with turbolinks taking over the page refresh. Should be an easy fix for someone more aux fait with JS.

Originally in #925

benortiz:
@BenJam hello! I'd be interested in picking this issue up and trying to find a solution.

andrew:
@benortiz Go for it 👍

benortiz:
Just wanted to give y'all an update with my progress so far. I was doing some digging to find the root of the problem and my current theory is that this is an iOS/Safari specific bug (please correct me if I'm wrong) related to sticky hover states. It looks like there was [_almost_](https://getbootstrap.com/docs/4.1/getting-started/browsers-devices/#sticky-hoverfocus-on-ios) a solution, but not any more. Still on the hunt for a solution.

pradyunsg:
I'm also seeing this on the current FF on Android 8.1.0.

andrew:
I wonder if we can just disable the hover/focus/active state on the button for small screens as a workaround?

PR #1092, opened 2018-10-12
Sync button is transparent for touch devices

benortiz:
On touch devices, the hover state is persisted instead of cleared when the user stops touching the element. I've used [CSS4 input mechanism media queries](https://caniuse.com/#feat=css-media-interaction) to restyle the button when the device does not support hovering. 

The funny thing about this approach is that you no longer get any feedback from pressing the button. I'm open to other solutions that keep this button feeling responsive. Also, I wasn't sure where in the CSS files this should live; I'm happy to move it wherever it should be. 

Fixes: https://github.com/octobox/octobox/issues/1036

<details>
<summary>Gif of solution</summary>
</img>
</details>


benortiz:
Yep, no problem! It did seem a little funny to have no indication. I'll see what I can do!

PR #1093, opened 2018-10-12
Status webhooks should update subects with combined status result

### LINKED ISSUE #1089: CI status isn't consistent with the one showed on GitHub pull requests listing.
 - [x] octobox.io
 - [ ] My instance

🐞 **Problem**

I have noticed several pull requests that are shown as "CI passed" in Octobox whereas on GitHub only one CI check is green and the others are pending.

🎯 **Goal**

The rule used by GitHub to aggregate status checks is:

- any failing check -> failing
- no failing and any pending -> pending
- all success -> success

It would be useful if Octobox could use the same rule. In the current situation, this status info is virtually useless to me.

andrew:
Fixes #1089

Had to do a bit of refactoring of the subject status syncing to be able to sync the combined status for a subject on receiving a `Status` webhook without a known notification/user. 

For GitHub App setups (https://octobox.io), this creates an [Installation Access Token](https://developer.github.com/v3/apps/#create-a-new-installation-token) to make the request, otherwise it uses the first user who appears to have access to the subject.

Also took the opportunity to deleted some unused subject methods and tests and reorganized the `SyncSubject` module.


prashuchaudhary:
👍

Issue #1133, opened 2018-10-19
Can we turn off the "Get more info for this Issue" prompt


gep13:
### What feature you'd like to see

I would like the ability to turn off the reminder/prompt to install the GitHub application:

![image](https://user-images.githubusercontent.com/1271146/47240536-80182080-d3e0-11e8-8fe1-9f99e3adbf1d.png)

## Motivation

I have already installed the GitHub App for all the repositories that I can, as there are some that I am not able to.  I don't need reminded about the GitHub Application, as I am already using, and loving it.

## Pitch

Simply put, it is a little bit annoying to see the reminder/prompt, when I already have it installed, and running, on other repositories.

hakusaro:
Not only that, but it's worth killing off the "install the app" banner in the user bar too.

hakusaro:
I sent a wip pr for this issue in #1136. If I've gone in the right direction I'll add some tests and see if I can get this upstreamed rapidly. :)

andrew:
I'd prefer to avoid adding more check boxes to the settings page.

I think we also need to take a different approach to solving https://github.com/octobox/octobox/issues/1133, just turning off all prompts doesn't really work to help get Octobox.io to become a sustainable open source project, which the GitHub app will be the main source of income.

The way I'd like to change the behavior of the popup would be to turn it from hover to click if the user has installed the GitHub app on any one or more repository that they have access to.

andrew:
https://github.com/octobox/octobox/pull/1183 will hide the message in the header once a user has authorized the GitHub app, even if they haven't installed it on any of their repos

PR #1136, opened 2018-10-20
WIP: Add setting to disable github app reminders

hakusaro:
This adds a user configurable setting that permits disabling the GitHub
app installation banner on the navbar as well as the tooltips in the
notification index.

It makes the UI less cluttered for people who won't install the app.

While "feature complete," I've marked this PR as WIP because I want positive confirmation that this is the way to go. I don't want to write tests for something if I'm not following a convention and everything needs to be reworked in the future. Can I get some feedback on if this is the right way to go, and whether or not I should go ahead and add tests for it?

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -105,6 +105,6 @@ class UsersController < ApplicationController
       params[:user][:personal_access_token] = nil
     end
 
-    params.require(:user).permit(:personal_access_token, :refresh_interval, :theme)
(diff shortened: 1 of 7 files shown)


andrew:
As mentioned in #1133 (comment), I'd prefer to not add a checkbox to hide this, we'll be experimenting a lot with the messaging around installing the GitHub App as it's a key part of making Octobox a sustainable open source project.

PR #1139, opened 2018-10-22
Revamp Docker Compose setup

chrisarcand:
## Description

'Revamp' is a strong word here; more like 'formalize'. Included are changes to allow the Dockerfile to create production-ready images and rework some of the Compose configuration and docs.

I've been doing a loooot of Docker work over the last several weeks, educating myself and finding the right balance of what Octobox should provide and what should be left to the user. The diff might look large and overwhelming, so let me highlight that the two largest changes in here are basically two lines of code: including production dependencies and precompiled assets in Docker images.

## Basic principles

Behind these changes are some specific intentions:

* The Dockerfile should build images that are ready for use _in production_, as would be expected of any such file. Note however that because people currently use the Dockerfile for development use - and because no maintainer here wants the overhead of maintaining separate Dockerfiles per-environment - the Dockerfile in this PR has everything needed for running in development or production.
* The provided `docker-compose.yml` file is for a _development_ environment. Writing your own configuration or overriding the file we provide with production values is entirely left to the user (and made a whole lot easier in this PR). 
* From a documentation perspective, we'll provide help for use of Docker with Docker Compose. There was a weird mix of Compose and vanilla Docker in the docs which is more confusing than anything else. As Compose is literally just a convenient way of specifying a configuration that you can run with normal Docker commands, using Docker without Compose (which is easy to do, seeing the Compose file) is left to the user. 

## Changes

With those principles in mind, here's the highlights of what this PR does:

* **Actually installs production dependencies in the Docker images(!)** Can't run Octobox with the build images without this.
* **Precompiles production assets in the image.** This also allows setting up a Docker volume for the common setup of serving assets outside the application server and follows the convention that Docker images should contain everything you need to run $the_thing. It also puts the time required for compilation at build time, not run time, which is nice. 

  Note that this brings the image size up to 802MB from 512MB. That might seem terrible, but consider that including production assets in images is done in other popular OSS Rails applications such as [Mastodon](https://github.com/tootsuite/mastodon/blob/c7e9f9ff1ed1def7f14f6ca4ac2836005eeefa47/Dockerfile#L84) (1.36GB!). The [Discourse image](https://github.com/discourse/discourse_docker/tree/master/image/base), which actually _doesn't_ include assets (compiles as a run time command) or even development dependencies, is 1.85GB!
* **Removes 'dev' Compose file.** This is pointless as the existing compose setup basically *is* a development configuration. Anyone running a production setup is going to want to override things anyway, thus, just keep the base file a development configuration and be done with it.
* **Ignores Compose override and production file for convenience**
* **Cosmetic changes** ($APP_ROOT, services ordering in Compose, etc)
* **Doc updates**, so many doc updates.

There might be a few more things to add later as I find them, but this is PR is getting large and is already a vast improvement, so I'm cutting myself off here.


andrew:
Good job on managing to push this to GitHub today! Will review tomorrow 👍

Dockerfile
```diff
@@ -22,20 +22,24 @@ RUN apk add --update \
     curl-dev \
  && rm -rf /var/cache/apk/* \
  && bundle config --global frozen 1 \
- && bundle install --without test production --jobs 2 \
+ && bundle install --without test --jobs 2 \
```
andrew: Thinking along similar lines, if we want people to be able to use the base docker image to do all the things that Octobox can do, should that also include run the test suite? For one thing that would mean we could integrate the Docker image into the CI pipeline

Dockerfile
```diff
@@ -22,20 +22,24 @@ RUN apk add --update \
...
 # ========================================================
 # Application layer
 
 # Copy application code
-COPY . /usr/src/app
+COPY . $APP_ROOT
+
+# Precompile assets for a production environment.
+# This is done to include assets in production images on Dockerhub.
+RUN RAILS_ENV=production bundle exec rake assets:precompile
```
andrew: Will this work if you’ve not got a production config (I.e database)? Would it still work without the RAILS_ENV?

andrew:
LGTM 👌

I’m not using the Docker image myself but I do try to build it wherever making some system related changes to check it mostly works.

andrew:
Looks like the travis webhooks never fired for this PR, might need an extra commit to kick it off

Dockerfile
chrisarcand: Maybe? That sounds right. I'll leave adding the test stuff as a separate change though (to be included on any other changes needed to be made by someone getting that all set up, if running the Docker image in the CI pipeline is a route you want to go).

Dockerfile
chrisarcand: > Will this work if you’ve not got a production config (I.e database)?

Yup I believe so.

> Would it still work without the RAILS_ENV?

AFAIK no. From what I understand assets are keyed per environment; in fact I've seen that only way to do something such as including precompiled assets for use in any environment is to literally repeat this command for all three envs 😧

samjacobclift:
I use the docker compose file for local development, this all looks good but did anyone manage to get live reloading working? i.e. if you edit a .css file or a ruby file can the docker file get rebuilt automatically without have to do docker-compose down/up cycle again?

chrisarcand:
I don't use Docker for local development so I'm not sure on that. Patches to do it are most welcome though, if it doesn't work!

andrew:
@samjacobclift if you run the app with `RAILS_ENV=development` then the rails app should reload itself without needing to start/stop any of the docker bits.

Issue #927, opened 2018-09-13
Dockerfile is development specific(?)


chrisarcand:
Are you experiencing this issue in octobox.io or your own instance?
 - [ ] octobox.io
 - [x] My instance

🐞 **Problem**

I'm completely new to Docker, so bear with me a bit and I realize it might just be my own lack of experience here.

I'm attempting to deploy an Octobox instance via [docker-compose in production](https://docs.docker.com/compose/production/). This would require me making my own custom compose yaml (which we might want to mention in the docs), which I've done - but I've hit a snag in that the Dockerfile we have [does not install gems for the production environment.](https://github.com/octobox/octobox/blob/88a6d926bdc7201035d1cf69137e3e6e7ae3b26d/Dockerfile#L22) That means gems used only in production cause the Rails initialization process to fail (in my case, lograge is missing.)

Should Dockerfiles be tied to the specific environment as ours currently is, or should we be installing dependencies for all potential environments? I'm assuming the latter, but I don't really know what I'm doing and I know other people are using Docker in "production like environments" (I'm wondering if they are all running in dev, actually, though)

🎯 **Goal**

I should be able to run Octobox in an actual production environment using Docker (and, hopefully, Compose) more easily.

💡 **Possible solutions**

I think we might just need to tweak the Dockerfile and leave slightly better instructions in the docs.

andrew:
@chrisarcand yeah we should definitely add back in the production gems, I seem to remember removing them to reduce the size of the docker image

chrisarcand:
Cool; Yeah I'm toying with this a bit and will have something soonish. I think a revamped set of compose yamls (generalize docker-compose.yml, add in overrides to be use in the dev yaml to be used [over the top of the general one](https://docs.docker.com/compose/extends/#example-use-case), and changes to the Dockerfile(s). 👈 We can either generalize the normal Dockerfile for all envs or we can also do a separate Dockerfile for production images to keep the size down (no dev/test gems, etc).

andrew:
You can tell that I’m not running it on Docker 😅

chrisarcand:
I'm setting up a new instance finally so I figured hey why not :) 🔜

rhuss:
I adapted the Dockerfile to include also production gems (i.e. removed that '--without production') but then run in another issue when running:

```
16:50:52 web.1    | bundler: failed to load command: puma (/usr/local/bundle/bin/puma)
[...]

Issue #667, opened 2018-06-25
Gradual memory leak with no plateau


davidcelis:
Are you experiencing this issue in octobox.io or your own instance?

* ~octobox.io~
* My instance

🐞 **Problem**

I run a personal Octobox instance on a DigitalOcean droplet (4GB of RAM) so that I can take advantage of the `FETCH_SUBJECTS` feature. On this instance, I also have personal access tokens enabled and am using one that I generated specifically for Octobox. My method for running it is via `docker-compose` (I configured all of the environment variables in a `docker-compose.yml` file).

After running the instance for a week or so, I started getting errors during syncing. The error was that the `pg` gem couldn't open connections to postgres due to a lack of RAM. Unfortunately, the server's RAM was so tied up that I couldn't even open an SSH connection; I had to log into DigitalOcean and perform a hard reboot 😰

After the hard reboot, the server and Octobox were up and running again. However, RAM continues to climb. This is a graph of memory usage since I rebooted the instance on the 20th:

![image](https://user-images.githubusercontent.com/36873/41869970-87438614-786f-11e8-9a77-93845fd74a19.png)

My assumption is that you haven't noticed a memory leak on octobox.io so, if that's the case, I also assume that it has to do with the `FETCH_SUBJECTS` feature.

🎯 **Goal**

Octobox, even with the experimental `FETCH_SUBJECTS` feature enabled, does not leak memory

💡 **Possible solutions**

A separate instance with memory debugging tools installed should probably be monitored for a short time to deduce the source of the leak

📋  **Steps to solve the problem**

Monitor an Octobox instance for memory leaks and figure out what resources are hanging around unnecessarily

chrisarcand:
> My assumption is that you haven't noticed a memory leak on octobox.io so, if that's the case, I also assume that it has to do with the FETCH_SUBJECTS feature.

Given the sheer amount of functionality that sits in that feature, probably - but more than that, I'm guessing that octobox.io and others don't have this problem because Heroku - IIRC dynos are restarted every 24 hours whether you want it or not.

davidcelis:
> IIRC dynos are restarted every 24 hours whether you want it or not.

Ohhhh yup, you're right https://devcenter.heroku.com/articles/dynos#restarting

andrew:
One other thing I've done to reduce memory usage on octobox.io is to switch to using jemalloc, might be worth trying that out before the more crude cronjob to restart it every 24 hours.

[...]

PR #1141, opened 2018-10-22
Add a prod-ready target to the Dockerfile

NiR-:
First of all, thank you a lot for that marvelous tool :grinning: 

This change leverage multi-stage build feature to create build targets
specific to development and production environments. Both targets are
based on a primal target named base. Packages and common gems are
installed in the base target and env-specific gems are installed in
their respective target. This should solve #927.

Using the production environment: enables PumaWorkerKiller (this should 
circumvent the issue described in #667), pull the logger level higher (so 
no more sql queries got dumped) and speed up the application by 
enabling some caching.

To minimize the image size, the `build-base` package is now installed
and removed every time `bundle` is run. This effectively reduce the size
of the dev build from 533MB to 364MB (for the dev target). And the
compressed size of the image on Docker Hub goes from 219MB to 159MB.

Also, note that Octobox now runs with uid 1000, to enforce better security 
practices.

To get this PR fully effective, I think you have to change a bit the way you 
deliver docker images: currently, it looks like all the branches are built for
dev environment. IMHO you should also (or instead), publish prod-ready 
images based on git tags. It'd also help operators determine what version
of Octobox is used.


NiR-:
Oh, I just realize @chrisarcand already opened a PR today to address some of the same issues this PR fixes.

The scope of his/their PR is wider than mine, however the introduction of the multi-stage feature in this PR, as well as the systematic deletion of `build-base`  help to reduce the size of the final image: while the final image from chrisarcand is ~800MB, mine is ~350MB (uncompressed). Thus, I think both are complementary.

andrew:
This will need a rebase now that 
https://github.com/octobox/octobox/pull/1139 has been merged

NiR-:
@andrew Done :)

Dockerfile
```diff
@@ -1,45 +1,81 @@
...
+
+# Copy application code
+COPY . $APP_ROOT/
+
+# Precompile assets for a production environment.
+# This is done to include assets in production images on Dockerhub.
+RUN RAILS_ENV=production rake assets:precompile \
+ && rm -rf /usr/src/app/tmp/cache/
+
+RUN RAILS_ENV=production rake assets:precompile \
```
andrew: Looks like this line might have been duplicated in the rebase?

andrew:
This is looking good to me, would be good for someone who's actually running Octobox with Docker to try this out before merging.

andrew:
I've pulled the docs fix out into a pr over here: https://github.com/octobox/octobox/pull/1168

chrisarcand:
Oh cool, thanks @NiR- ! I was reading about multistage stuff last week and figured something like this could be utilized as a followup to my recent PR, which you saw. Excellent!

I'll need to find a bit of time to sift through this and try it out; it's on my list to look at before the end of this week. Thanks for your patience!

Dockerfile
NiR-: Oh indeed, that was a rebase mistake. It's now fixed.

Dockerfile
```diff
@@ -1,45 +1,78 @@
...
 # * Generate the docs
 # * Make files OpenShift conformant
 RUN RAILS_ENV=development bin/rails api_docs:generate \
- && chgrp -R 0 $APP_ROOT \
+ && chown -R 1000:0 $APP_ROOT \
  && chmod -R g=u $APP_ROOT
 
-# Startup
-CMD ["bin/docker-start"]
+USER 1000
```
chrisarcand: From the docs... 

> The USER instruction sets the user name (or UID) and optionally the user group (or GID) to use when running the image and for any RUN, CMD and ENTRYPOINT instructions that follow it in the Dockerfile.

Doesn't that mean this should be before all the other commands in each section?

docker-compose-dev.yml
```diff
@@ -0,0 +1,25 @@
+version: '3.4'
```
chrisarcand: Why is this file added? An old artifact from before I did my own PR where the separate dev compose file was removed? Is this necessary?

chrisarcand: (to combine my feedback of 'we should just provide production images', if we do that I believe the target added in this file should just be edited in the existing compose file and this file removed)

chrisarcand:
Followup: I just looked and it seems that you can't actually specify separate build targets to automatically build on Docker Hub, furthering the thought that we shouldn't bother providing anything but production images. 👍

docker-compose-dev.yml
NiR-: Indeed, it's an old artifact, my bad.

> the target added in this file should just be edited in the existing compose file

If developers have to update their `docker-compose.yml` file to run dev versions, this will not be really practical as it'll be marked as an "unstage change" by git. This might introduce too much friction. However, we can use an env var, defined in `.env`, so people can easily switch to any env.

nesl247: My recommendation is to have `docker-compose.yml` only include what is useful in EVERY environment, and add a `docker-compose.override.yml.dist` file which is designed for environment specific purposes, with defaults for development. Users then copy it to `docker-compose.override.yml` and make any changes necessary. You add the `.dist` to `.gitignore` and then it won't show up as an unstaged change.

jakubgs:
Any news on this?

Dockerfile
NiR-: Unfortunately this would mean running apk with user 1000 and that won't work. Since `bundle install` commands are all surrounded by apk commands, the only place where we can set the `USER` is at the end of each target.

However, AFAIK bundler and RubyGems does not have the same security issue than, for instance, npm: there's no external code run during installation phase, so I believe there's no reason for the build environment to be compromised (from that vector). In the end, it's only there to not run Octobox as a root user, as this is considered a bad practice (see [here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user)).

andrew: Rubygems does actually allow for arbitrary code execution during installation via the option to build native extensions: http://blog.costan.us/2008/11/post-install-post-update-scripts-for.html although it requires jumping through more hoops than with npm.

NiR-: Ha too bad :/ I'm not really sure what to do then. I tried to find out articles about "dockerizing" ruby apps, but almost none talks about `USER` instruction. The only one I found using this instruction, does so after installing the dependencies (see [here](https://blog.kontena.io/dockerizing-ruby-application#exampleapplication)).

Do you think this is not an issue in the end, or should I do the required changes to not run bundler as root (that would mean we couldn't install/deinstall build-tools before/after running bundler)?

NiR-:
I updated the PR to include some more changes:
* The .dockerignore file now works like a whitelist instead of the usual blacklist mode. That should prevent anyone building images from sources to inadvertently include files with credentials (like `.env.backup` or anything like that).
* Docs has been updated to contain instructions to set up prod instances, with an example `docker-compose.yml` and a `.env` exhaustively listing config parameters.
* The `app` container in dev env now has a volume to bind-mount the whole project. This is better since `cache_classes` is turned to false in that env.

andrew:
This has some conflicts but I'd really like to parts of it merged, would be good to test what it's like upgrading an existing docker install with this new image and config.

andrew:
Fixed up the conflicts here, going to look at how to get docker hub to build the production version on push.

andrew:
Pushing it to docker hub didn't really work 🤔

NiR-:
Oh, what's the issue?

andrew:
```
app_1                                | `/` is not writable.
app_1                                | Bundler will use `/tmp/bundler/home/unknown' as your home directory temporarily.
app_1                                | rake aborted!
app_1                                | LoadError: cannot load such file -- rubocop/rake_task
app_1                                | /usr/local/bundle/gems/bootsnap-1.4.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:22:in `require'
app_1                                | /usr/local/bundle/gems/bootsnap-1.4.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:22:in `block in require_with_bootsnap_lfi'
app_1                                | /usr/local/bundle/gems/bootsnap-1.4.4/lib/bootsnap/load_path_cache/loaded_features_index.rb:89:in `register'
app_1                                | /usr/local/bundle/gems/bootsnap-1.4.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:21:in `require_with_bootsnap_lfi'
app_1                                | /usr/local/bundle/gems/bootsnap-1.4.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:40:in `rescue in require'
app_1                                | /usr/local/bundle/gems/bootsnap-1.4.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:26:in `require'
app_1                                | /usr/local/bundle/gems/activesupport-5.2.3/lib/active_support/dependencies.rb:291:in `block in require'
app_1                                | /usr/local/bundle/gems/activesupport-5.2.3/lib/active_support/dependencies.rb:257:in `load_dependency'
app_1                                | /usr/local/bundle/gems/activesupport-5.2.3/lib/active_support/dependencies.rb:291:in `require'
app_1                                | /usr/src/app/rakefile:26:in `<top (required)>'
app_1                                | /usr/local/bundle/gems/rake-12.3.2/exe/rake:27:in `<top (required)>'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/cli/exec.rb:74:in `load'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/cli/exec.rb:74:in `kernel_load'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/cli/exec.rb:28:in `run'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/cli.rb:463:in `exec'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/vendor/thor/lib/thor.rb:387:in `dispatch'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/cli.rb:27:in `dispatch'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/vendor/thor/lib/thor/base.rb:466:in `start'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/cli.rb:18:in `start'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/exe/bundle:30:in `block in <top (required)>'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/lib/bundler/friendly_errors.rb:124:in `with_friendly_errors'
app_1                                | /usr/local/bundle/gems/bundler-2.0.1/exe/bundle:22:in `<top (required)>'
app_1                                | /usr/local/bundle/bin/bundle:23:in `load'
app_1                                | /usr/local/bundle/bin/bundle:23:in `<main>'
app_1                                | 
app_1                                | Caused by:
app_1                                | Bootsnap::LoadPathCache::FallbackScan: 
app_1                                | 
app_1                                | (See full trace by running task with --trace)
octobox_app_1 exited with code 1
```

sjbarag:
Just want to say I'd _love_ to have this merged upstream for everyone to use!  @NiR- are you still able to drive this PR?

docs/INSTALLATION.md
```diff
@@ -78,7 +78,7 @@ Octobox uses [`encrypted_attr`](https://github.com/attr-encrypted/attr_encrypted
 
 Therefore to install and launch Octobox, you must provide a 32 byte encryption key as the env var `OCTOBOX_ATTRIBUTE_ENCRYPTION_KEY`
 
-Protip: To generate a key, you can use `bin/rails secret | cut -c1-32`
+Protip: To generate a key, you can use `bin/rails secret | cut -c1-32`. With docker: `docker run --rm -it akerouanton/octobox:latest bin/rails secret | cut -c1-32`.
```
mikemimik: Should this be `octobox/octobox:latest`?

docs/INSTALLATION.md
```diff
@@ -174,24 +166,34 @@ that, you can override the `docker-compose.yml` configuration by adding a
...
-The `docker-compose.yml` file provided is for a _development_ configuration;
-there are are a number of things you'll want to configure differently for
-production use, like setting the Rails application for production and setting
-up a reverse proxy such as Apache or Nginx to serve static assets. You can use the
-`docker-compose.yml` file as an example to write your own or simply override
-the existing configuration with `docker-compose.override.yml`. Both the
-override file and `docker-compose.production.yml` are gitignored.
+The `docker-compose.yml` file provided at the root of the project is for a
+_development_ configuration only. However, you can download the example
+`docker-compose.yml` file from [here](https://raw.githubusercontent.com/octobox/octobox/master/docs/examples/docker-compose.yml).
```
mikemimik: This url link is no longer valid. I don't believe there is an example of a _production_ `docker-compose.yml` file anymore in this repo.

chrisarcand:
While we're here, when retesting this again, the Redis image needs to be updated to 4.0 for the inclusion of Rails 6.

glitchcrab:
I too would love to see this landed, it would iron out any of the niggles I currently have with Octobox. More than happy to help test if needed too.

PR #1156, opened 2018-10-27
show issue/PR numbers together with title

### LINKED ISSUE #1109: Also show issue/PR numbers together with title
## Motivation

It would be nice if issue & PR numbers would somehow also be visible together with the title, perhaps in a darker font so it stands out less than the title (should be in a small column left of the title imho).

Personally, I feel that I frequently filter by PR number mentally (i.e. long-running issues/PRs stand out because I start recognizing the number after a while).

In addition, in EasyBuild we frequently rely on PR numbers to test contributions, via the `eb --from-pr` support we have (cfr. https://easybuild.readthedocs.io/en/latest/Integration_with_GitHub.html#using-easyconfigs-from-pull-requests-from-pr).

## Pitch

Also showing issue/PR number would make it easier to "map" between the octobox overview & the GitHub interface, imho.

chaserx:
Resolves #1109 

**updated**: normal font-weight for the number specifically. 

screen shot 2018-11-11 at 10 15 04 am


docker-compose.yml
```diff
@@ -11,7 +11,7 @@
 #   actively develop Octobox, you can override this file by adding the following to
 #   a docker-compose.override.yml:
 #
-#   version '3'
+#   version: '3'
```
chaserx: i guess this could be in a separate PR, but this tripped me up when setting up

docs/INSTALLATION.md
```diff
@@ -167,7 +167,7 @@ that, you can override the `docker-compose.yml` configuration by adding a
 `docker-compose.override.yml` with the following:
 
 ```yaml
-version '3'
+version: '3'
```
chaserx: same as above

boegel:
@chaserx Thanks a lot for looking into this! Is there any way to make the title stand out a bit more, for example by using a different (dimmer) color for the issue/PR number?

chaserx:
@boegel yeah, that's possible. i'll see what I can do.

chaserx:
@boegel I've updated the screengrab in the description above to show normal font-weight of just the number. thoughts?

andrew:
@chaserx what happens if the repo name has a number in it? Like https://github.com/24pullrequests/24pullrequests

chaserx:
@andrew well... it'll match `24` twice. That's a problem.

chaserx:
@andrew we can get more specific by adding the end of line symbol with something like `\d+$`. 

Example:
http://rubular.com/r/7pmogcVF0i

chaserx:
@andrew it probably wouldn't hurt if i write a couple of tests for this helper method. 😃

andrew:
@chaserx maybe move the number method to the Notification class and unit test it from there?

chaserx:
Moved and tested. Happy to add other test cases.

boegel:
@chaserx Looks better! :+1:

docker-compose.yml
andrew: Bounced into a new pr: https://github.com/octobox/octobox/pull/1166

docs/INSTALLATION.md
andrew: Bounced into a new pr: https://github.com/octobox/octobox/pull/1166

chaserx:
conflicts resolved.

app/helpers/notifications_helper.rb
```diff
@@ -289,6 +289,14 @@ def sidebar_notification_status(status)
     octicon(NOTIFICATION_STATUS_OCTICON[status], height: 16, class: "sidebar-icon #{status}")
   end
 
+  def subject_with_number(notification)
+    if notification.subject_type == 'Issue' || 'PullRequest'
+      "<span class=\"notification-number\">##{notification.subject_number}</span> #{notification.subject_title}".html_safe
```
andrew: Having the HTML concatenated inline feels kinda bad here, especially as it requires using `.html_save` which makes me worry about XSS attacks.

Perhaps instead, in the view we can just conditionally render the number and it's span?

chaserx: ah! yes, I can fix that up this evening.

andrew:
@chaserx brilliant, thanks for updating, shipping it shortly 🎉

PR #1183, opened 2018-11-05
Only show the Github App install prompt if the user hasn't authorized the app

andrew:
Related to https://github.com/octobox/octobox/issues/1133


andrew:
As mentioned in https://github.com/octobox/octobox/issues/1133#issuecomment-434330045, I'd prefer to not add a checkbox to hide this, we'll be experimenting a lot with the messaging around installing the GitHub App as it's a key part of making Octobox a sustainable open source project.

PR #1221, opened 2018-11-14
Allow notifications horizontal scrolling

HipsterBrown:
To fix the case where notification actions, like archive or mute, are hidden by the notification thread section, allow horizontal scrolling for overflow content.

Example: 

The quick actions are hidden from access due to the thread area. 

screen shot 2018-11-14 at 11 13 48 am

### DIFF
--- a/app/assets/stylesheets/_layout.scss
+++ b/app/assets/stylesheets/_layout.scss
@@ -50,7 +50,7 @@ body {
 .flex-main {
   flex: 1;
   min-width: 0;
-  overflow-x: hidden;
+  overflow-x: scroll;
   padding-left: 0;
 }
 


BenJam:
OOI what screen size are you running there? I'd prefer to make sure that the thread window sizes properly (it should go to the responsive version at twbs' 'large' breakpoint.

BenJam:
okay so I can't get this to do the same thing, because the text in my notification titles isn't long enough to force the td to explode the table element:

screenshot 2018-11-15 at 15 30 48

this is a problem that re-occurs at other perfect breakpoints and long subject names too...

HipsterBrown:
I've fixed this for now by shortening my window to just below the breakpoint so the thread panel does not always show. That's a better experience for me anyway because it's an empty view for the majority of the time I have Octobox open. Thanks for looking into this. I can close the PR as the fix needs to be more comprehensive. 😄

PR #1248, opened 2018-11-18
Fix links in header dropdown menu

bastilian:
When notifications are open, links will have a different root and
need to be made absolute in order to avoid that.


andrew:
Good catch!

PR #1262, opened 2018-11-20
Use correct token for repos not in the Github App

### LINKED ISSUE #1261: PR 'Metadata' not being pulled with Personal Access Token, when Github App is enabled
 - [ ] octobox.io
 - [x] My instance

🐞 **Problem**

I am running Octobox on my own instance, and am using a Personal Access token for the oauth, so that it can access the Repos that are owned by the organization.
I also have the FETCH_SUBJECT=true flag, and it was working beautifully before this weekend.

This weekend I 'upgraded' to using a GitHub App. It only has access to my Personal repos, and for those it is working perfectly!

However all my repos under 'The Org' aren't getting any of the extra meta-data they used to. The one that is the most important to me is the PR status icon, which is currently just the white icon.

Is it possible/does Ocotobox already fall back to trying the FETCH_SUBJECT if there is a GitHub app installed, that simply doesn't have permissions to access all my repos?

Does this sound like a code issue to you? And if so do you have an idea where the offending code might live? I am more than happy to take a look and try to get a PR out if this is a real bug!

Also let me know if you would like this as a separate issue, and I will move it over!

🎯 **Goal**

Make it so the Personal Access token is used for the relevant requests. This will allow people to use Octobox, without getting explicit permission from an Org to install the Github App.

💡 **Possible solutions**

I'm more than willing to help out on the fix if someone is willing to point me in the right directions!

coreyja:
Fixes #1261

This attempts to make sure when we are fetching the Subject information, we are using the 'correct' access token. 'correct' in this context follows the following logic:
  - If the User (and instance has them enabled) has a Personal Access Token, we want to use that token
  - Elsif the the notification comes from a repo that has the github app installed, use the app token
  - else fallback to the oauth access token

There are also some slight refactors that I thought made things more clear, but feel free to let me know if you think otherwise!

I currently don't have any automated tests on any of this. I quickly scanned the existing tests and didn't see a place where this type of logic was tested.
While this is no substitute for automatic testing, I am thinking I will deploy this branch to my instance tonight.

I also have a few questions I am going to ask inline.

Thanks so much for your time, and all the work you do on this project!

- [x] Test Cases

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -92,7 +92,7 @@ class User < ApplicationRecord
   end
 
   def personal_access_token_client
-    @personal_access_token_client ||= Octokit::Client.new(access_token: personal_access_token, auto_paginate: true) if personal_access_tokens_enabled?
+    @personal_access_token_client ||= Octokit::Client.new(access_token: personal_access_token, auto_paginate: true) if personal_access_token_enabled?
   end
 
   def access_token_client
@@ -123,7 +123,7 @@ class User < ApplicationRecord
     "#{'*' * 32}#{personal_access_token.slice(-8..-1)}"
   end
 
-  def personal_access_tokens_enabled?
+  def personal_access_token_enabled?
     Octobox.personal_access_tokens_enabled? && personal_access_token.present?
   end
 
--- a/lib/octobox/notifications/sync_subject.rb
+++ b/lib/octobox/notifications/sync_subject.rb
@@ -26,7 +26,7 @@ module Octobox
       end
 
       def github_client
-        if user.personal_access_tokens_enabled?
+        if user.personal_access_token_enabled?
           user.personal_access_token_client
         elsif app_installation.present?
           user.app_installation_client


lib/octobox/notifications/sync_subject.rb
```diff
@@ -28,7 +28,7 @@ def update_subject_in_foreground(force = false)
       private
 
       def download_subject
-        user.subject_client.get(subject_url)
+        github_client.get(subject_url)
```
coreyja: I believe this is the actual root issue I am attempting to fix. Before my changes `user.subject_client` would always use the `subject_token` if the GithubApp is installed.

After this change it will only use this token if the repository is accessible by the Github App

app/models/subject.rb
```diff
@@ -139,9 +139,9 @@ def download_comments
 
   def github_client
     if app_installation.present?
-      Octobox.installation_client(app_installation.github_id)
+      app_installation.github_client
```
coreyja: This is just using a method that already exists!

app/models/subject.rb
```diff
@@ -139,9 +139,9 @@ def download_comments
...
     else
-      users.first&.subject_client
+      users.first&.github_client
```
coreyja: This is a change in logic, but I believe if we already know the `app_installation` is not present for this Repo, we should NOT try to use the app access token

app/models/notification.rb
```diff
@@ -172,4 +174,12 @@ def subject_number
   def display_thread?
     Octobox.include_comments? && subjectable? && subject.present? && user.display_comments?
   end
+
+  def github_client
+    if app_installation.present?
+      user.app_installation_client
```
coreyja: So here I am using `app_installation_client`. This matches most closely what was happening here: https://github.com/octobox/octobox/pull/1262/files#diff-cf94d05b646cd6dc6622869c79a2c4dbL31

Where it would use this token if the App was installed for the User. 

I call this out becuase it is different from the `#github_client` method that we have on `Subject`.
I think the root of my question is, What is the difference between 

```
Octokit::Client.new(access_token: user.app_token, auto_paginate: true)
```

and

```
Octobox.installation_client(self.github_id)
```


I _think_ that this usage here, using `user.app_installation_client` is correct! Since this is trying to fetch User scoped data, so I think we need a user scoped token. The `installation_client` seems to be in a User-less context.
So if that is correct, I am not sure if that means the usage in `Subject` is correct...

app/models/subject.rb
coreyja: While this is existing I am not sure if I believe this is using the correct client! See my comment up in `Notification#github_client` for more context

app/models/user.rb
```diff
@@ -93,12 +93,8 @@ def github_client
     @github_client
   end
 
-  def subject_client
-    Octokit::Client.new(access_token: subject_token, auto_paginate: true)
-  end
-
-  def subject_token
-    app_token || effective_access_token
+  def app_installation_client
```
coreyja: The ONLY usage of this is now in `#sync_app_installation_access` in this model.

I renamed it in an attempt to make things more clear, but I am not sure if this worked lol

coreyja:
Anecdotally, this does seem to fix the issue I was having in my instance as described by #1261 🎉

app/models/subject.rb
andrew: 👍 for the deduplication

app/models/subject.rb
andrew: 👍

app/models/notification.rb
```diff
@@ -172,4 +174,12 @@ def subject_number
   def display_thread?
     Octobox.include_comments? && subjectable? && subject.present? && user.display_comments?
   end
+
+  def github_client
```
andrew: I'd probably move this method into `lib/octobox/notifications/sync_subject.rb` as that looks like it's the only place it gets used

app/models/user.rb
andrew: I think this makes sense, as it's only used in `sync_app_installation_access` then we can move this method `private` to the user class

app/models/notification.rb
```diff
@@ -172,4 +174,12 @@ def subject_number
...
+    if app_installation.present?
```
andrew: I think we'll need to add `app_installation` to the eager loading over here: https://github.com/octobox/octobox/blob/master/app/services/download_service.rb#L67 to avoid n+1 sql queries when syncing notifications.

app/models/notification.rb
andrew: Correct, the `user.app_token` is only different from a token generated for the App in that it has more context around the user making the request, which in this case doesn't matter.

lib/octobox/notifications/sync_subject.rb
andrew: 👍

andrew:
When it comes to testing, I'd be tempted to test that `Notification.github_client` has the correct token in a number of situations:

- Personal access token is present on the user
  - when the notification's repo has a github app installed (expect personal access token)
  - and when it doesn't (expect personal access token)
- Personal access token isn't present on the user
  - when the notification's repo has a github app installed (expect app token)
  - and when it doesn't (expect access token)

coreyja:
@andrew Those test cases sound great, I will look into getting those added! One point of clarification.

> Personal access token is present on the user
when the notification's repo has a github app installed (expect personal access token)

In this case do you think we want to use the Personal Access Token, over the app token? That is not what this PR does currently, but I can definitely change it if that is the logic you think is correct!

andrew:
In general I believe that the personal access token has much greater privileges than the app could (which is why they aren’t enabled on octobox.io), the only reason I can see to use the GitHub app setup with a personal access token is for the webhooks.

app/models/user.rb
coreyja: I was wrong about this, I am using it in `Notification#github_client` so I think it needs to stay public

lib/octobox/notifications/sync_subject.rb
```diff
@@ -25,10 +25,20 @@ def update_subject_in_foreground(force = false)
...
       end
 
+      def github_client
+        if user.personal_access_tokens_enabled?
+          user.github_client
+        elsif app_installation.present?
+          user.app_installation_client
+        else
+          user.github_client
+        end
```
coreyja: This methods reads a bit weird, but the logic is correct. I am very open to suggestions about how to make this code clearer.

`github_client` picks between the Personal Access Token and the 'normal' Access Token, so when we call it on line 30 we expect to be using the Personal Access Token, while when we call it on line 34 we are expecting to use the 'normal' Access Token

coreyja: I pulled apart the User methods more so we don't have to rely on just `github_client` here!

coreyja:
Bump! I think this is ready for review!

app/models/user.rb
```diff
@@ -114,17 +115,21 @@ def effective_refresh_interval
   end
 
   def effective_access_token
```
andrew: Is this method even used anymore?

PR #1263, opened 2018-11-20
Add index on github_id for repositories table

andrew:
Speeds up a number of lookups in background jobs


andrew:
cc @jules2689 another index, although this one is on a much smaller table

PR #1271, opened 2018-11-20
Show a spinner when loading a thread view

andrew:
Useful on slower connections so you know something is happening.

Looks like this when loading (but with the icon spinning):

screen shot 2018-11-20 at 14 50 20


BenJam:
lgtm

Issue #1273, opened 2018-11-20
Help people understand the difference between Sign in and Install the app


BenJam:
Following from https://mobile.twitter.com/_alanshaw/status/1064485257901236224 

Some ideas:
* move the 'sign in' button to top left. which makes it more like a log in journey/
* make the 'install the GH app' more about the extra 'enhanced' notifications piece, adding further detail (and images) to the 'enhanced' bit OR
* Put some detail in the 'get more info about this notification' dialog, incl. images.

PR #1280, opened 2018-11-21
Disambiguate 'installing' from 'signing in'

BenJam:
* move sign in to top right
* add '/org' to pricing
* add 'liven up notifications' section
* remove overflow on `.container`

Addresses #1273 

Kinda unsure about the structure now. Seems like the 'feature' cards should be re-framed in terms of user-pain a la 'Don't lose track'. 

Open to comments:

screencapture-localhost-3000-2018-11-21-14_41_57


BenJam:
Cut down:

screenshot 2018-11-21 at 16 40 09

Issue #1285, opened 2018-11-22
"Mark as read" when all selected should uncheck "Select all" checkbox when done


bastilian:
Are you experiencing this issue in octobox.io or your own instance?
 - [ ] octobox.io
 - [X] My instance

🐞 **Problem**

When all notifications in a list are selected with the "Select All" checkbox and "Mark as read" is used it does not uncheck it after it completed the action.

💡 **Possible solutions**

"Select All" should be unchecked or items that stay in the list should not be unchecked.

📋  **Steps to solve the problem**

* Select All via the checkbox
* Click "Mark as Read"
* "Select All" stays checked, while items marked as read get unselected

![octobox_selected_items_issue](https://user-images.githubusercontent.com/7757/48895862-66865080-ee46-11e8-94d3-32dfca4ebc75.gif)

andrew:
@bastilian are you using firefox? I've noticed that selected checkboxes seem to persist in firefox across refreshes in a number of places.

andrew:
Also 💯 for including a gif in your bug report 👏

bastilian:
@andrew Yeah, the issue is/was in Firefox. I tried #1289, but that did not solve the issue. However adding an `uncheckAll` function to trigger after "Mark as Read" is done does the trick. :) I put it into #1293 and it too has a nice GIF!

It would be nice though if any selection would persist, but atm. most of the times only one action is done on a particular selection.

PR #1293, opened 2018-11-24
Uncheck all selected notifications after "Mark as Read" is done

bastilian:
When "Select All" is used with "Mark as Read" it will only uncheck list items,
but leave "Select All" checked.

This will use "Select All" to uncheck All items and fixes the issue in #1285.

kapture 2018-11-24 at 17 32 11


andrew:
Nice fix, thanks!

PR #1294, opened 2018-11-24
Run Rubocop in CI

coreyja:
I realized that Rubocop was configured in this project, but wasn't set up to run in CI

This PR integrates the Rubocop Rake tasks, and also hooks it up to run in Travis. The first commit will most likely fail, since there are Rubocop violations on Master right now. I will follow that up with probably a combination of a todo file and some fixes


.rubocop.yml
```diff
@@ -157,6 +160,7 @@ Lint/UnusedMethodArgument:
   Description: 'Checks for unused method arguments.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#underscore-unused-vars'
   Enabled: true
+  AutoCorrect: true
```
coreyja: I had to add this to get these to get --auto-correct to work on these rules. I think it has something to do with the usage of `DisabledByDefault` but I'm not 100% sure

PR #1318, opened 2018-12-01
Add 10 to pagination intervals

timhoffm:
When viewing octobox on small screens like an iPad, 20 notfications do not fit on the screen. Thus you'll always have to scroll to see the lower part of the items list and the navigation buttons.

This PR adds a new item for just showing 10 notifications per page, so that everything is visible without scrolling.


andrew:
👍

Issue #1304, opened 2018-11-28
Add different bg colors on notifications while navigating in organisation


benoittgt:
### What feature you'd like to see
I would love to have a better view when I am looking for notifications of a different project in a same organisation. 

## Pitch

Here are two proposals:

![screenshot-2018-11-28-at-10 08 472](https://user-images.githubusercontent.com/8417720/49141221-6aa6e800-f2f6-11e8-82f0-cc4e18d81686.gif)

or less visible

![screenshot-2018-11-28-at-10 08 473](https://user-images.githubusercontent.com/8417720/49141238-71cdf600-f2f6-11e8-95c2-7246c687af3f.gif)

Thanks 😄

BenJam:
Hey @benoittgt thanks for the feature request. I wonder whether you could elaborate a little on 

> a better view when I am looking for notifications of a different project

Some initial interpretations might be that better means that it's clearer which notifications correspond to which projects. In which case we could solve the problem here my ordering by project > datetime as much as colouring. As is this could be similar to using colours in excel: good in small doses but not that useful when looking at 100 rows. 

Either way I think digging into what the core problem you have here would be incredibly useful for us so please do share any thoughts you have.

benoittgt:
> ordering by project > datetime as much as colouring

I think it can be a solution but it is still less clear for me than with color. I think when you have 100 rows you will probably use the dedicated repo view. For me color is helpfull in case like this one

![screenshot 2019-01-02 at 15 00 40](https://user-images.githubusercontent.com/8417720/50595068-431be000-0e9f-11e9-94af-b579846abbdf.png)

Few projects but without a lot of issues or PR. The "owner" view is powerfull but quite messy when you are focusing on repo column. Like a news feed color or avatar help you to split between *owners* quickly.

I am not sure I properly answer to your comment. This is not a problem if you think this is not relevant and want to close the issue (sorry @RebeccaAR).

Thanks for the follow up

BenJam:
@benoittgt thansk for your comments. I wonder whether colouring the owner/repo combination a la the reasons and labels would love this problem.

It seems like your commetns about avatars are a slightly different issue. are you frequently looking for an 'owner' of a particularl subject (issue, PR etc?). We used to have that, but it was removed some time ago. @andrew might be able to lend some context as to why.

andrew:
If you have the [GitHub app](https://github.com/marketplace/octobox/) installed then the owner column may help with this, without needing to add extra colours to the UI:

[...]

PR #1319, opened 2018-12-01
Solution for different colours for different repos

RebeccaAR:
Solution for colour coding different repos/projects 

#1304


RebeccaAR:
image

here you go :)

andrew:
cc @benoittgt 👆

benoittgt:
Thanks a lot!

BenJam:
Firstly, thanks for putting this together @RebeccaAR. And apologies for the delay in getting around to this, December was mosly given over to holiday and getting the report for sustainoss.org drafted. 

As is I'm not really a fan of this solution as it leaves colour to chance and can end up with some quite horrid results:

image

I hear the issue from @benoittgt though. I just wonder whether there's another approach here. I'll pick up in #1304.

BenJam:
@RebeccaAR I've presented a simple alternative in https://github.com/octobox/octobox/issues/1304#issuecomment-453543414 if you'd like to have another quick go. Else I may take it up Monday.

PR #1345, opened 2018-12-07
Add rack-mini-profiler and other performance gems

tarebyte:
I was looking into an N+1 in our instance of Octobox and thought it'd be nice a nice addition for future use.

This adds the following gems

- rack-mini-profiler
- memory_profiler
- flamegraph
- stackprof

It also adds the pop up in the bottom right hand corner always in dev and for admins in production.

✨ Thanks ✨


andrew:
I'm reluctant to have these gems present all the time for all users when they are likely only to be used for infrequent performance tuning, the memory usage of sidekiq is already topping out the average heroku dyno as it is.

I wonder if we can keep it as a branch or option that can be deployed for performance investigation purposes?

andrew:
Going to close this for now but will keep the branch around.

PR #1351, opened 2018-12-12
Added dump.rdb to the .gitignore file

jhsu802701:
The dump.rdb file is created every time I run the test suite and should be added to .gitignore.


andrew:
Thanks!

Issue #1284, opened 2018-11-21
Highlight security vulnerabilities when they're in the inbox


BenJam:
In a similar vein to #1267 we may wish to highlight (in red) when a new security vulnerability is present when `unread(this.notificiation)` and `inbox:true`:

<img width="252" alt="screenshot 2018-11-21 at 16 55 21" src="https://user-images.githubusercontent.com/158833/48856428-4440f400-edae-11e8-950e-6fb48acf7dc4.png">

andrew:
👀

gep13:
@BenJam ooo, I like it!

Rodrigue-K:
I'm going to work in it.

andrew:
This is in progress here: https://github.com/octobox/octobox/pull/1352

PR #1352, opened 2018-12-13
Issue 1284 highlight alerts

Rodrigue-K:
This commit fixes the #1284 issue by adding an unread-alert class for "RepositoryVulnerabilityAlert'  to the notifications_helper. However I don't think querying the database from notification helper is the best solution. Would love some feedback/ideas on improving this solution.

`def sidebar_filter_link(active:, param:, value:, count: nil, except: nil, link_class: nil, path_params: nil, title: nil)
    css_class = 'nav-item'
    css_class += ' active' if active
    css_class += ' unread-alert' if unread_alerts?(value)`

`def unread_alerts?(type)
    type == 'RepositoryVulnerabilityAlert' &&
      Notification.reorder(nil).
                   where(unread: true, subject_type: "RepositoryVulnerabilityAlert").count > 0
  end`

### DIFF
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -232,7 +232,11 @@ module NotificationsHelper
   end
 
   def sidebar_filter_link(active:, param:, value:, count: nil, except: nil, link_class: nil, path_params: nil, title: nil)
-    content_tag :li, class: (active ? 'nav-item active' : 'nav-item'), title: title do
+    css_class = 'nav-item'
+    css_class += ' active' if active
+    css_class += ' unread-alert' if unread_alerts?(value)
+
+    content_tag :li, class: css_class, title: title do
       active = (active && not_repo_in_active_org(param))
       path_params ||= filtered_params(param => (active ? nil : value)).except(except)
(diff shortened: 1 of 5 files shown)


app/helpers/notifications_helper.rb
```diff
@@ -232,7 +232,11 @@ def repo_filter_link(param, repo_name, count)
   end
 
   def sidebar_filter_link(active:, param:, value:, count: nil, except: nil, link_class: nil, path_params: nil, title: nil)
-    content_tag :li, class: (active ? 'nav-item active' : 'nav-item'), title: title do
+    css_class = 'nav-item'
+    css_class += ' active' if active
+    css_class += ' unread-alert' if unread_alerts?(value)
```
andrew: Perhaps here instead of making a database query we could add a css class to name each sidebar_filter_link, for example by using the `param` and `value` arguments so we get a class like `type-RepositoryVulnerabilityAlert`. 

You could change line 237 to:

    css_class += " #{param}-#{value}"

BenJam:
would love to see this in 👍

BenJam:
I'll pick this up if @Rodrigue-K is busy elsewhere :)

PR #1354, opened 2018-12-14
Added the bundler-audit and gemsurance gems

jhsu802701:
Because using gemsurance adds tmp directory files that are flagged by RuboCop, I updated .rubocop.yml to exclude the tmp directory in its analysis.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,6 +3,8 @@ inherit_from: .rubocop_todo.yml
 AllCops:
   DisabledByDefault: true
   TargetRubyVersion: 2.5
+  Exclude:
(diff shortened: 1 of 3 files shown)


jhsu802701:
I updated .rubocop.yml to omit the vendor directory, because RuboCop flagged files in it for offenses in the Travis environment.

andrew:
Similar to #1355, I'm not a fan of adding these gems for one-off tasks, many people run octobox in development mode at the moment and this will unnecessarily increase their memory usage. 

We use https://dependabot.com to keep our gems up-to-date which includes security notifications already.

PR #1355, opened 2018-12-14
Added the railroady and rails-erd gems

jhsu802701:


andrew:
I'm not a fan of adding these gems for one-off tasks, many people run octobox in development mode at the moment and this will unnecessarily increase their memory usage. 

The railroady gem actually has a command line interface that doesn't require it to be added to the Gemfile if you need it.

PR #1374, opened 2018-12-17
save selected notifications after marking as read

### LINKED ISSUE #1348: Item selection is clear after an action on it
 - [x] My instance

🐞 **Problem**
It looks like now the select item selector is clear after one action is done in the items selected and another action could be performed in the same items.
It used not to be that way. the selector used to remember the items after perform an action on them.

**Sceanrio:**
* Select few unreal items
* mark them as read
**Issue:**
* The selection is clear

evexoio:
fix #1348

gif

Sorry for the dark overlay, its night and I have to protect my eyes 😍


andrew:
Thanks for looking into this 💝

PR #1426, opened 2019-01-04
Added debug box

jhsu802701:
I added a debug box to the footer for the development environment only.


BenJam:
thanks for the contribution @jhsu802701. Could you include a screen-grab next time you make a visual change? Bit easier to review. For instance: 

screenshot 2019-01-07 at 10 49 07

Our footer is a little weird, it's not really a footer. We may readdress this when we move thread view out of beta. So this might have to hang until then.

PR #1433, opened 2019-01-07
Require paid plan for notifications of private repos on octobox.io

andrew:
A big change we're making this year to the hosted version of Octobox is that notifications on private repositories will require a paid plan, either on individual repositories with an organization plan ($100/month for unlimited users) or for an individual user via a personal plan ($10/month for each user).

screen shot 2019-01-07 at 13 32 55

I'm still experimenting with exactly how to display this change whilst still enabling users to know that they have notifications on private repositories and to be able to hide/archive those notifications.


BenJam:
🚢

PR #1439, opened 2019-01-09
Only attempt to load comments in thread view if subject is present

andrew:
Fixes a rare exception when a thread view is loaded when there's no subject to display.


BenJam:
👍

PR #1467, opened 2019-01-15
Rolling fixes to thread revamp

BenJam:
* show GH ids not octo ids
* open GH in new window
* show 'no description' for blank subjects/comments
* shrink subject length to accommodate comment count on responsive table
* fix logic of last (shown) comment header


BenJam:
this is an empty comment.

Issue #1470, opened 2019-01-15
Header behaviour on iOS 11 broken


BenJam:
It's a while ago now but iOS 11 is playing havoc with some of the header content, which is particularly annoying on threads. 

Useful blog post on rectifying the situation: https://ayogo.com/blog/ios11-viewport/

PR #1472, opened 2019-01-15
iOS  headers and sticky fixes

BenJam:
Addresses #1470 

WIP


andrew:
I think I may have fixed the underlying issue in https://github.com/octobox/octobox/pull/1714

PR #1484, opened 2019-01-16
Update thread when receiving notification updates and new comments

BenJam:
I might end up working this over with comment updates later today/tomorrow. 

* moves header to separate partial
* sends both the notification and the thread header down the wire
* renders the notification and the thread if open
* moves existing hooks to notifications channel
* adds new comments channel for subject currently viewed
* unsubscribes from subject feeds when changing thread


BenJam:
I'll just leave this for a mo.

BenJam:
ready for review, apologies for horrendous commit-log!

PR #1486, opened 2019-01-16
Only download notifications that have been updated since last_synced_at

andrew:
Whilst investigating speeding up some of the background jobs, I noticed that when calling `#sync_notifications` on a user would result in attempted updates to notifications that hadn't changed.

I initially sped that up by skipping some work on notifications that hadn't changed (https://github.com/octobox/octobox/pull/1464 and https://github.com/octobox/octobox/pull/1468) but I was still left with a feeling that the [`DownloadService`](https://github.com/octobox/octobox/blob/d3dc604920a3af36f2579c86fd0b6cf10041e2a5/app/services/download_service.rb) was pulling down more notifications that required. 

After doing some more poking around in the production console, I found that even if you ran `sync_notifications_in_foreground` twice in quick succession, the second time would still download and process many notifications for some users, which are effectively no-ops as they were all synced 
the first time and nothing had changed.

This lead me to look at the `fetch_read_notifications` method:
```ruby
def fetch_read_notifications
  oldest_unread = user.notifications.unread(true).newest.select(:updated_at).last
  if oldest_unread && oldest_unread.updated_at.respond_to?(:iso8601)
    headers = {cache_control: %w(no-store no-cache)}
    since = oldest_unread.updated_at - 1.second
    notifications = fetch_notifications(params: {all: true, since: since.iso8601, headers: headers})
    process_notifications(notifications)
  end
end
```

This method has been in Octobox for a long time, over 2 years infact: https://github.com/octobox/octobox/pull/227

What it appears to do is look for the oldest existing notification for a user and use the `updated_at` timestamp to limit the number of notifications fetched from the GitHub API, but not just read but unread too. 

The weird thing is that this method is called along side `fetch_unread_notifications`, which uses the `last_synced_at` timestamp of a user if it's present to only get notifications that are new or updated since the last time we synced.

Turns out that for some users, they have a really old, unread notifications, which then causes `fetch_read_notifications` to ask for all the notifications since `1.year.ago` (for example).

Out of curiosity I commented out the call to `fetch_read_notifications` in https://github.com/octobox/octobox/blob/d3dc604920a3af36f2579c86fd0b6cf10041e2a5/app/services/download_service.rb#L28 to see which tests would break, but none broke!

That lead me down a refactoring path that ended up as this pull request, where we use a modified version of `fetch_unread_notifications` that gets both read and unread notifications that have changed since the last time the user synced their notifications (or all of them for new users).

I can't see any issue with merging this but changing code that's been running quietly in production for two years made me question if the tests are really testing the right things but it looks like all the bases are covered there too.

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -140,11 +140,11 @@ class Notification < ApplicationRecord
     end
   end
 
-  def update_from_api_response(api_response, unarchive: false)
+  def update_from_api_response(api_response)
     attrs = Notification.attributes_from_api_response(api_response)
     self.attributes = attrs
     self.archived = false if archived.nil? # fixup existing records where archived is nil
-    unarchive_if_updated if unarchive
+    unarchive_if_updated
     if changed?
       save(touch: false)
       update_repository(api_response)
--- a/app/services/download_service.rb
+++ b/app/services/download_service.rb
@@ -23,13 +23,7 @@ class DownloadService
 
   def download
     timestamp = Time.current
-
-    if user.last_synced_at
-      fetch_read_notifications
-      fetch_unread_notifications
-    else
-      new_user_fetch
-    end
+    fetch_new_notifications
     user.update_column(:last_synced_at, timestamp)
   end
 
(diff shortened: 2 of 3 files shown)


andrew:
Going to merge this and deploy to https://octobox.io for production testing, can always roll it back but after re-reading all the notification tests I'm now feeling pretty confident that this change won't make any different in the end user experience.

andrew:
The savings in memory and cpu for sidekiq from this are looking 👌 
screen shot 2019-01-16 at 17 30 13

chrisarcand:
Wow, this is awesome! Nice work.

PR #1521, opened 2019-01-23
Always add read:org to scopes when limiting access

davidcelis:
Currently, we don't add `read:org` to the default OAuth scopes when
Octobox is also connected to a GitHub App. However, this leads to a bug
where access is denied to _private_ members of the organization. I was
able to manually allow access by defining the `GITHUB_SCOPE` variable
but it would be good to get an official fix in :smile:


andrew:
Thanks!

PR #1523, opened 2019-01-24
Post comments directly to a subject from Octobox

BenJam:
Adds a form to the bottom of the thread view to post a comment on that subject.

* adds form, controller, model methods et al to accept, process and render the comment
* adds comment worker, working a new comment queue (could add this to current user queue)
* escapes shortcuts if we're focussed on the comment box

I am in two minds as to whether to bother adding a JS action on this as turbolinks is doing a good job of re-rendering. But will happily add if needed. 

Future extensions of this might add previews for markdown, file uploading etc. this is just a starter to build upon. 

Will need some work to #1484 to ensure that comments aren't re-rendered when receiving the hook back from GH which I'll add there this avo.


BenJam:
Builds and tests fine locally, can't see why Travis is throwing a wobbly...

app/assets/javascripts/channels/comments.js
```diff
@@ -24,7 +24,7 @@ if ($("meta[name='push_notifications']").length >0) {
         channel: "CommentsChannel",
         notification: $(location).attr('href').split('/').pop()},{
         received: function(data){
-          if ($('#notification-thread').attr('data-id') == data.subject_id){
+          if ($('#notification-thread').attr('data-id') == data.subject_id && !$("#comment-"+data.comment_id).length){
             $('.discussion-thread').append(data.comment_html);
```
andrew: Looks like you've got the same code here twice, but the existing comment check only in the second block? Can you pull the logic out into a function to reduce the duplication?

app/controllers/notifications_controller.rb
```diff
@@ -87,7 +87,7 @@ def show
     if @notification.subject && @notification.subject.commentable?
       comments_loaded = 5
       @comments = @notification.subject.comments.order('created_at DESC').limit(comments_loaded).reverse
-      @comments_left_to_load = @notification.subject.comment_count - comments_loaded > 0 ? @notification.subject.comment_count - comments_loaded : 0
+      @comments_left_to_load = @notification.subject.comments.count - comments_loaded > 0 ? @notification.subject.comments.count - comments_loaded : 0
```
andrew: Is this change really required? It causes 2 extra sql queries

app/controllers/notifications_controller.rb
```diff
@@ -98,16 +98,17 @@ def show
...
   def expand_comments
 
-    scope = notifications_for_presentation.newest
+    scope = original_scope = notifications_for_presentation.newest
     scope = load_and_count_notifications(scope) unless request.xhr?
 
     ids = scope.pluck(:id)
     position = ids.index(params[:id].to_i)
-    @notification = current_user.notifications.find(params[:id])
+    @notification = original_scope.find(params[:id])
```
andrew: Looks like this change is an accidental revert from a merge?

app/controllers/notifications_controller.rb
```diff
@@ -98,16 +98,17 @@ def show
 
   def expand_comments
 
-    scope = notifications_for_presentation.newest
+    scope = original_scope = notifications_for_presentation.newest
```
andrew: Looks like this change is an accidental revert from a merge?

app/controllers/notifications_controller.rb
```diff
@@ -98,16 +98,17 @@ def show
...
 
     ids = scope.pluck(:id)
     position = ids.index(params[:id].to_i)
-    @notification = current_user.notifications.find(params[:id])
+    @notification = original_scope.find(params[:id])
     @previous = ids[position-1] unless position.nil? || position-1 < 0
     @next = ids[position+1] unless position.nil? || position+1 > ids.length
 
     @comments_left_to_load = 0
+    @more_comments = false
```
andrew: Looks like this change is an accidental revert from a merge?

app/controllers/notifications_controller.rb
```diff
@@ -122,6 +123,16 @@ def expand_comments
     end
   end
 
+  def comment
+    subject = @current_user.notifications.find(params[:id]).subject
```
andrew: Shouldn't this be `current_user` instead of `@current_user`?

app/models/subject.rb
```diff
@@ -124,6 +124,21 @@ def update_comments
     end
   end
 
+  def comment(user, comment_body)
+    return if comment_body.nil?
```
andrew: Maybe check for empty strings here? `comment_body.empty?`

app/workers/comment_worker.rb
```diff
@@ -0,0 +1,29 @@
+class CommentWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: :comments, unique: :until_and_while_executing
+
+  def perform(comment_id, user_id, subject_id)
+  	comment = Comment.find_by_id(comment_id)
```
andrew: Indentation is a little off here

app/workers/comment_worker.rb
```diff
@@ -0,0 +1,29 @@
...
+    user = User.find_by_id(user_id)
+    subject = Subject.find_by_id(subject_id)
+    if comment && user && subject && subject.commentable?
+      begin
+        subject.comment_on_github(comment, user) 
+      rescue Octokit::NotFound, Octokit::Unauthorized, Octokit::Forbidden
+        comment.try(:destroy)
+      rescue Octokit::BadGateway, Octokit::ServerError, Octokit::ServiceUnavailable => exception
```
andrew: This rescue should be removed so that sidekiq retries it

test/factories/comment.rb
```diff
@@ -1,5 +1,6 @@
 FactoryBot.define do
 	factory :comment do 
+		subject
```
andrew: Indentation is off here

test/workers/comment_worker_test.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class CommentWorkerTest < ActiveSupport::TestCase
+  setup do
+  	@user = create(:user)
```
andrew: Indentation is off here

app/controllers/notifications_controller.rb
BenJam: possibly yeah, I had a bit of a sketchy merge couple fo days back.

test/factories/comment.rb
BenJam: weird, I thought Rubo would pick these up!?

app/assets/javascripts/channels/comments.js
BenJam: exiting comment being shown would only happen if we had just posted a comment (which kick a turbolink load). App.comments is re-initialised on each of these, so if I attach it there the code is in two places again. If I attach it to Octobox then the code is in another place which seems a bit weird. Not sure where else to drop it.

BenJam:
made those changes (other than the JS, which I'm happy to take an approach on).

app/assets/javascripts/channels/comments.js
andrew: I'd prefer to reduce the duplication to avoid future copy-pasta bugs

BenJam: not questioning that strategy, just not sure where to attach that code.

app/workers/comment_worker.rb
BenJam: I think I removed this correctly, but please check.

andrew:
This whole block of js is almost exactly the same (except for `$(location)` or `$(id)`), can it all be extracted?

```js
      if(App.comments){
        App.comments.unsubscribe();
      }
      App.comments = App.cable.subscriptions.create({
        channel: "CommentsChannel",
        notification: $(location).attr('href').split('/').pop()},{
        received: function(data){
          if ($('#notification-thread').attr('data-id') == data.subject_id && !$("#comment-"+data.comment_id).length){
            $('.discussion-thread').append(data.comment_html);
          }
          addCommentToThread(data.subject_id, data.comment_id, data.comment_html)
        }
      });
```

andrew:
Just one thing left to fix up, the usage of `@current_user` which should be `current_user`

BenJam:
done

BenJam:
> `FETCH_SUBJECT` enabled on installation
I don't think the thread would be viewable if it wasn't.

> `write_issue` currently not saved in db, need to add a new field to `app_installations`
requires a permission update too no? is that okay?

> subject's repo is open source (must use the current users access_token not app_token)
do we have something for checking this?

andrew:
> > `FETCH_SUBJECT` enabled on installation
> 
> I don't think the thread would be viewable if it wasn't.

Installations that are setup as GitHub apps don't need `FETCH_SUBJECT` enabled to pull in subjects for some repositories (like https://octobox.io) and since https://github.com/octobox/octobox/pull/1409 all installations can pull in open source subjects without `FETCH_SUBJECT` being enabled.
 
> > `write_issue` currently not saved in db, need to add a new field to `app_installations`
> 
> requires a permission update too no? is that okay?

Yeah that will require a permission update for https://octobox.io, as well as a documentation update for anyone else running it as a github app here: https://github.com/octobox/octobox/blob/master/docs/INSTALLATION.md#running-octobox-as-a-github-app

The existing code that stores the permissions for each app is over here: https://github.com/octobox/octobox/blob/master/app/models/app_installation.rb#L88

> > subject's repo is open source (must use the current users `access_token` not `app_token`)
> 
> do we have something for checking this?

You can do `subject.repository.open_source?` as defined over here: https://github.com/octobox/octobox/blob/master/app/models/repository.rb#L15

app/models/user.rb
```diff
@@ -143,4 +143,13 @@ def sync_app_installation_access
     removed_permissions = app_installation_permissions.reject{|ep| app_installation_ids.include?(ep.app_installation_id) }
     removed_permissions.each(&:destroy)
   end
+
+  def can_comment?(subject)
+    return subject.commentable? && subject.repository.commentable? && #is the subject/repo commentable? (checkes for write issues)
+      personal_access_token_enabled? || #do we have a personal access token?
+      Octobox.fetch_subject? || # are we running with repo scope?      
+      subject.repository.app_installation_id || #there is an app installation on the repo (we already checked for write permission)
```
andrew: You say `we already checked for write permission` but I can't see where this check is happening?

app/models/subject.rb
```diff
@@ -123,6 +123,21 @@ def update_comments
...
 
+  def comment(user, comment_body)
+    return if comment_body.nil? || comment_body.empty?
+    comment = comments.create(author: user.github_login, body: comment_body)
+    CommentWorker.perform_async_if_configured(comment.id, user.id, self.id)
+  end
+
+  def comment_on_github(comment, user)
+    return if comment.body.empty?
+    remote_comment = user.github_client.post "#{url}/comments", {body: comment.body}
```
andrew: The `github_client` use here needs to be swapped out with the particular kind of client depending on the permissions setup for that repository. Namely, if the repository is private and has the github app installed, then the `app_installation_client` needs to be used instead of the regular `github_client`.

app/controllers/notifications_controller.rb
```diff
@@ -122,6 +123,16 @@ def expand_comments
     end
   end
 
+  def comment
+    subject = current_user.notifications.find(params[:id]).subject
+    subject.comment(current_user, params[:comment][:body]) if subject.commentable?
```
andrew: We should check if the user `can_comment?` here before attempting it, and then redirect with an error if they cannot.

app/controllers/notifications_controller.rb
```diff
@@ -87,7 +87,8 @@ def show
     if @notification.subject && @notification.subject.commentable?
       comments_loaded = 5
       @comments = @notification.subject.comments.order('created_at DESC').limit(comments_loaded).reverse
-      @comments_left_to_load = @notification.subject.comment_count - comments_loaded > 0 ? @notification.subject.comment_count - comments_loaded : 0
+      @comments_left_to_load = @notification.subject.comment_count - comments_loaded 
+      @comments_left_to_load = 0 if  @comments_left_to_load < 0
```
andrew:
```suggestion
      @comments_left_to_load = 0 if @comments_left_to_load < 0
```

app/models/app_installation.rb
```diff
@@ -66,6 +66,14 @@ def github_client
     Octobox.installation_client(self.github_id)
   end
 
+  def write_issues?
+    permission_issues ? permission_issues == 'write' : false
+  end
+
+  def read_issues?
+    permission_issues ? permission_issues == 'read' : false
```
andrew: I think this will return false if `permission_issues` is `'write'`, would be good to get a test for this, I believe we can just check to see if there's any value in `permission_issues` instead,

app/models/repository.rb
```diff
@@ -20,6 +20,11 @@ def github_app_installed?
     app_installation_id.present?
   end
 
+  def commentable?
+    return app_installation.write_issues? && display_subject? unless app_installation.nil?
```
andrew: I think this logic may need expanding to add in conditions about `FETCH_SUBJECT` and personal access tokens, which will allow commenting even if there's an app installation with only `read_issues`

app/models/user.rb
andrew: > You say `we already checked for write permission` but I can't see where this check is happening?

Ok, it looks like that's happening in `subject.repository.commentable?` but the `||` are shortcutting that check after the first one, would be good to either add brackets around the groups of conditions or break this out into separate checks so it's easier to read.

test/controllers/notifications_controller_test.rb
```diff
@@ -655,6 +655,7 @@ class NotificationsControllerTest < ActionDispatch::IntegrationTest
 
   test 'search results can filter by label' do
     sign_in_as(@user)
+    Subject.delete_all
```
andrew: This line shouldn't be needed now
```suggestion

```

test/models/user_test.rb
```diff
@@ -142,4 +142,48 @@ def assert_error_present(model_object, error)
...
+
+    @token_user = create(:token_user)
+    
+    repository = create(:repository, private: true)
+    installation = create(:app_installation, repositories: [repository])
+    subject = create(:subject, repository: repository)
+    
+    assert @user.can_comment?(subject)
+  end
+
```
andrew: Couple more test cases:

- `user can comment on a subject when FETCH_SUBJECT enabled`
- `user cannot comment on a private subject with an app installation without an app_token`

app/assets/javascripts/channels/comments.js
```diff
@@ -1,34 +1,26 @@
-if ($("meta[name='push_notifications']").length >0) {
+var subscribeToComments = function(){
 
-  $(document).on('click', '.thread-link', function(){
-    if(App.comments){
-      App.comments.unsubscribe();
-    }
-    App.comments = App.cable.subscriptions.create({
-      channel: "CommentsChannel",
-      notification: $(this).attr('href').split('/').pop()},{
```
andrew: There might be a race condition here, changing `$(this).attr('href').split('/').pop()` to `$(location).attr('href').split('/').pop()` requires that the code is called after the url has changed from `/` to `/notifications/123`, I'm not sure how the order of on click callbacks is defined in jquery, how do we make sure this on click happens after the turbolinks happens after https://github.com/octobox/octobox/blob/master/app/assets/javascripts/application.js#L69

app/models/app_installation.rb
BenJam: I think `permission_issues` is 'read' by default (at least it looked that way on the live DB for octobox's repo. Will add a test.

BenJam: just ran this in irb and it seemed okay: maybe I am getting the type collapsing wrong?

```
irb(main):002:0> permission_issues = 'write'
=> "write"
irb(main):004:0> permission_issues ? permission_issues == 'read' : false
=> false
irb(main):005:0> permission_issues ? permission_issues == 'write' : false
=> true
```

app/controllers/notifications_controller.rb
```diff
@@ -124,12 +124,19 @@ def expand_comments
...
+    if current_user.can_comment?(subject)
+      subject.comment(current_user, params[:comment][:body]) if subject.commentable?
+      if request.xhr?
+        render partial: "notifications/comments", locals:{comments: subject.comments.last}, layout: false
+      else
+        redirect_back fallback_location: notification_path
+      end
     else
-      redirect_back fallback_location: notification_path
+      flash[:error] = 'Could not update your account'
```
andrew: copy pasta error

app/models/repository.rb
```diff
@@ -21,8 +21,8 @@ def github_app_installed?
   end
 
   def commentable?
-    return app_installation.write_issues? && display_subject? unless app_installation.nil?
-    display_subject?
+    return  (display_subject? && app_installation.write_issues?) || 
+      current_user.personal_access_token_enabled?
```
andrew: `current_user` isn't available in this context

test/models/user_test.rb
BenJam: > `* user cannot comment on a private subject with an app installation without an app_token`

ISWYDT

app/models/repository.rb
```diff
@@ -20,9 +20,12 @@ def github_app_installed?
...
 
-  def commentable?
-    return  (display_subject? && app_installation.write_issues?) || 
-      current_user.personal_access_token_enabled?
+  def commentable?(user)
+    if app_installation.nil?
+        return user.personal_access_token_enabled? || Octobox.fetch_subject?
+    end
+    (display_subject? && app_installation.write_issues?) ||
+        user.personal_access_token_enabled? || Octobox.fetch_subject?
```
andrew: 👍

app/models/app_installation.rb
andrew: `permission_issues ? permission_issues == 'read' : false` => false is the problem, if the `permission_issues` field is `'write'`, then both `read_issues?` and `write_issues?` should return true.

Currently `read_issues?` returns false if the `permission_issues` field is `'write'`.

BenJam:
picking up the post-token stuff later today after a couple of days working on boardgames 👍

BenJam:
wondering why the heck I have a test failing on a dependency bump merge...

BenJam:
> wondering why the heck I have a test failing on a dependency bump merge... 

Looks like this erroring out on a reset back to b4e9d57

BenJam:
Right then. If I comment out the necessary sections I find that commenting out L:154 in `models/user` makes the `user cannot comment on a private subject without an app installation token` pass. So somewhere the test is being run against a user that has personal access tokens enabled. But even if I set `stub_personal_access_tokens_enabled(value: false)` as the first line in that test, it still fails on that test.

BenJam:
> Would be good to have a test for this

not sure how to test this. capturing the post from `subject.comment_on_github` and checking against the users key would be my go to. But I'm not sure how to do that.

app/models/user.rb
```diff
@@ -104,6 +104,11 @@ def access_token_client
     @access_token_client ||= Octokit::Client.new(access_token: access_token, auto_paginate: true) if access_token.present?
   end
 
+  def comment_client(comment)
+    return app_installation_client if app_token.present? && comment.subject.repository.private?
```
andrew: Shouldn't this be `comment.subject.repository.commentable?`?

BenJam:
Good to go.

app/models/user.rb
BenJam: should the method give you a client to try to post a comment, even if the installation doesn't have write issues. No. Probably not. But it does kinda conflate two concerns. Meh. I'll sort it out.

andrew: It mostly needs to check if the app is installed on the repository, otherwise any requests with the app_installation_client for that repository will fail

andrew:
We could also do with updating the documentation, mentioning the permissions change at least.

BenJam:
I have this running fine from my personal instance over at https://octobox-benjam.herokuapp.com/ fine. I'll get on with the docs etc.

andrew:
> I have this running fine from my personal instance over at https://octobox-benjam.herokuapp.com/ fine.

Have you tried setting up an instance as a github app?

andrew:
Would be great to do cmd+enter to submit a comment straight from the keyboard

andrew:
Finally merged 🎉  commenting will be enabled on octobox.io soon, to enable on your own instance, upgrade the issues permissions of your GitHub app from "read" to "write".

PR #1536, opened 2019-01-26
Only show google analytics privacy policy section if it's enabled.

andrew:
We have Google analytics in our privacy policy but it's not actually enabled at the moment.


app/views/pages/privacy.html.erb
```diff
@@ -51,31 +51,33 @@
...
     </p>
 
     <h3>Cookies</h3>
     <p>
-      Cookies are small amounts of text stored on your computer's web browser that can be written by websites and read by websites. They can also be shared between websites. We use cookies to store some information about your current session, this enables us to do things like keep you logged into the website between pages. Cookies are also used by third-party providers like Google:
+      Cookies are small amounts of text stored on your computer's web browser that can be written by websites and read by websites. They can also be shared between websites. We use cookies to store some information about your current session, this enables us to do things like keep you logged into the website between pages. <% if Rails.application.secrets.google_analytics_id.present? %>Cookies are also used by third-party providers like Google:<% end %>
```
BenJam: might want to triple check no session info from cookies sent on to bugsnag. Might be safer to keep this clause in completely and not name a third party provider in the case that we add something in the future adn forget to change this (or add something like Danger/Peril)

andrew: Updated to always say "third parties" instead of specifically (and conditionally) Google.

PR #1547, opened 2019-01-29
Update Open Collective donator > subscriber rake task

BenJam:
Updated rake task in line with yesterday's pricing changes 

* Fixes regex
* Adds 'Open Collective Organisation' and 'Open Collective Individual' plans to donors who have donated over 100 or between 10 and 99 dollars in the last month

Note that this /will/ try to assign an organisation plan to those who elected to donate 100 or more in a given month. I think we're okay not havign to handle this for the moment.

Feel free to commit over the top with the 'right' plan names in the DB @andrew.


andrew:
I've created those two subscription plan names now, any previous OC customers will be grandfathered onto the old plan

PR #1559, opened 2019-02-01
Add dynamic theme-color content

Jackymancs4:
Hello,

I use a dynamic theme for Firefox, and it mainly uses the `theme-color` meta tag to set the accent color for the browser, but even when in dark mode it always defaults to white.

Beside my specific use case, I know about other applications that works in similar ways (e.g. Vivaldi browser).

The solution proposed in the PR is not very scalable, but if you don't plan to support other default themes or custom ones, it should get the job done.


andrew:
Thanks!

Issue #1222, opened 2018-11-14
Implement synchronous donation journey


BenJam:
Looks like we can create something on par with the GH marketplace journey using https://github.com/opencollective/opencollective/issues/1402

Would need to:

- [ ] log the user in
- [ ] check the current, repeating donations for the user
- [ ] show the user their current donations on an interstitial with a button to:
- [ ] redirect to opencollective and create a new repeating donation for the additional amount
- [ ] check the donation amount on redirect and:
- [ ] allocate an OC plan for the month
- [ ] drop them into the main notifications page with a status update about their news subs.

only thing this misses is one-off donations within the month. I have a preference to send people to a 'repeating' option by default for obvious reasons.

BenJam:
> we should make a specific page that loads their transactions, says thanks and shows if they've donated the required amount to enable the private repository enhancements.

We get to set the transaction amount and type (single, repeat) so we can take the existing, repeat subs and 'top up' the amount. I would say ignore the recent,. single donations as they will end up with us dropping users at the end of the month when that single donation rolls off.

andrew:
Steps:

1. Redirect user to open collective url `https://opencollective.com/octobox/donate/100/month/individual?redirect=https://octobox.io/opencollective/callback`
2. User gets redirected back to `https://octobox.io/opencollective/callback?transactionid=12345` with a transaction id
3. We load the transaction via their api using a privately generated `api_key`: `https://api.opencollective.com/v1/collectives/octobox/transactions/12345?api_key=xxxxx`
4. We then update/create the correct plan and subscription purchase for that user

andrew:
Step 1 currently 404ing on open collective, reported here: https://github.com/opencollective/opencollective/issues/1402#issuecomment-458508463

PR #1563, opened 2019-02-07
Synchronous open collective donation journey

andrew:
Closes https://github.com/octobox/octobox/issues/1222

Current implementation only works for individual $10/month donations due to a limitation of the OC payments page.

~Haven't been able to test this yet as I don't seem to be able to make an API key that can load transactions for Octobox via the open collective website, will email them to find out.~

Other things that would be nice to do, depending on how receptive OC is to each suggestion:

- Allow users to switch between different plans on the payment UI page
- If user selects an org, then after redirect ask them to choose which github org the donation should apply to
- transactionid should be more random than an incrementing number, current incrementing id implementation allows someone to loop through every possible transactionid.
- embeddable donation form from open collective that we can display on our site
- webhooks for when a user cancels/upgrades their donation amount
- always have users redirected to https://octobox.io/opencollective even if they started the donation journey on https://opencollective.com/octobox

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -24,6 +24,10 @@ class ApplicationController < ActionController::Base
 
   private
 
+  def check_octobox_io
+    redirect_to '/422', flash: {error: 'This page is only available on https://octobox.io'} unless Octobox.io?
+  end
+
   def add_user_info_to_bugsnag(notification)
     return unless logged_in?
 
--- a/app/controllers/open_collective_controller.rb
+++ b/app/controllers/open_collective_controller.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 class OpenCollectiveController < ApplicationController
+  before_action :check_octobox_io
+
   def callback
     transaction_id = params[:transactionid]
     transaction = Octobox::OpenCollective.load_transaction(transaction_id)
--- a/app/controllers/open_collective_controller.rb
+++ b/app/controllers/open_collective_controller.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+class OpenCollectiveController < ApplicationController
+  def callback
+    transaction_id = params[:transactionid]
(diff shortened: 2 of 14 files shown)


andrew:
There appears to be a bug with looking up transactions by ID on opencollective's side at the moment which is blocking this, am in contact with someone on the OC slack about it.

znarf:
@andrew The API endpoint should be fixed tomorrow.

znarf:
@andrew Shipped.

andrew:
@znarf thanks!

BenJam:
I'm randomly getting an error locally on `bin/rails test test/controllers/open_collective_controller_test.rb:14` WRT stubbing out that request. Even though I can see the stub right there... 

Similar on `test/controllers/open_collective_controller_test.rb:29` and `test/controllers/open_collective_controller_test.rb:41`

BenJam:
In general looks good for immediate needs.

Does the donation get made straight away or on the next donation date? it's a little weird that OC appears to run one huge donation script on the 1st of every month and the 1st of each year. Just notices as I was going through the flow locally:

screenshot 2019-02-08 at 14 28 24

Some edge case thoughts: might be worth extending this to take into account existing donations (edge case) and might be worth ensuring that the end date for a subscription updates if someone donated part way through an existing subscription.

znarf:
@BenJam Just to confirm the behavior: 1 immediate full charge, next charge the 1st of next month. It's currently being discussed here on OC side. https://github.com/opencollective/opencollective/issues/1215

What would be the alternatives?
- Charge only a partial subscription if you're in the middle of the month (Feb 8th -> ~ 6 euros)
- Trigger the next charge on the same day of next month (Feb 8th -> March 8th)
- Skip the immediate donation and wait for the 1st of next month

andrew:
> I'm randomly getting an error locally on `bin/rails test test/controllers/open_collective_controller_test.rb:14` WRT stubbing out that request. Even though I can see the stub right there...

Can you post the error message?

PR #1578, opened 2019-02-15
Disable downgrading open collective subscribers for now

andrew:
Differences between data in different opencollective API endpoints causing paying users to be disabled incorrectly.

Should fix @Mr0grog's account being repeatedly downgraded.


Mr0grog:
Woohoo, thank you! 🎉 🙇

PR #1596, opened 2019-02-18
Activated the channel and validator tests

jhsu802701:
This pull request activates the channel and validator tests, which are not being run under the skip_visuals banner.  Activating the lib and services tests led to many errors.  We need to fix those tests before we can activate them.


jhsu802701:
Run the test suite in Travis again, because the one that fails is the notifications controller test that is prone to intermittent failures.

PR #1616, opened 2019-02-22
Quick, temp fix to lower open collective donor support requests

BenJam:
I had hoped I'd get more time today to look at this but, it ran away with me.

I know that making error message html_safe is not a great thing to do but: in order to stop people from sending support requests I propose adding this until something more permanent can be done.


BenJam:
Ideally we would be able to add something to Open Collective's FAQ page that highlights the need to add your GitHub username to your Open Collective account after donating.

andrew:
We only need their github username if they don't get sent back to us via the sync flow

BenJam:
I think my rake task (which needs updating) would remove them so I'll put that on the todo list...

PR #1626, opened 2019-02-24
Fix login redirect loop in Safari

andrew:
If you visit /login when already logged in, go directly to the homepage rather than back to to oauth.


BenJam:
🎢

PR #1631, opened 2019-02-25
Add default searches to new users

BenJam:
Adds three default 'pinned searches' to new user accounts:

- Archivable: any closed or merged issues or PRs that are not yet in the archive
- Mergeable: any open pull request that is currently passing CI 
- My PRs: any PR that is in your inbox for which you were the author

The aim here is to get every new user _aware_ of pinned searches and to help onboard them quickly by removing all the chaff (archivable) focussing on quick wins (mergeable) and then getting focussed on their own work (my prs).

I'm not sure how best to roll this across the 12k users we have on Octobox.io. Some help there would be 👌

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -165,9 +165,8 @@ class User < ApplicationRecord
   end
 
   def create_default_pinned_searches 
-    pinned_searches.new(query: 'state:closed,merged archived:false', name: 'Archivable')
-    pinned_searches.new(query: 'type:pull_request state:open status:success', name: 'Mergeable')
-    pinned_searches.new(query: "type:pull_request author:#{github_login} inbox:true", name: 'My PRs')
-    save
+    pinned_searches.create(query: 'state:closed,merged archived:false', name: 'Archivable')
+    pinned_searches.create(query: 'type:pull_request state:open status:success', name: 'Mergeable')
+    pinned_searches.create(query: "type:pull_request author:#{github_login} inbox:true", name: 'My PRs')
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -161,4 +163,11 @@ def can_comment?(subject)
     return true if github_app_authorized? && subject.repository.commentable?
     return false
   end
+
+  def create_default_pinned_searches 
+    pinned_searches.new(query: 'state:closed,merged archived:false', name: 'Archivable')
+    pinned_searches.new(query: 'type:pull_request state:open status:success', name: 'Mergeable')
+    pinned_searches.new(query: "type:pull_request author:#{github_login} inbox:true", name: 'My PRs')
+    save
```
andrew: This just calls save on the user, not the previously initialized objects

andrew:
For existing users, perhaps we can create a rake task that checks if they have any pinned searches, and if not calls `create_default_pinned_searches` on them, otherwise we can just run it in the production console.

app/models/user.rb
```diff
@@ -160,4 +162,10 @@ def can_comment?(subject)
     return true if github_app_authorized? && subject.repository.commentable?
     return false
   end
+
+  def create_default_pinned_searches 
+    pinned_searches.create(query: 'state:closed,merged archived:false', name: 'Archivable')
+    pinned_searches.create(query: 'type:pull_request state:open status:success', name: 'Mergeable')
```
pradyunsg: Should this also include archived:false?

andrew: Fixed, good spot!

BenJam: ++ thanks @pradyunsg

PR #1633, opened 2019-02-25
Remove commenting from users without a token with repo scope

### LINKED ISSUE #1632: Reply/Comment on GitHub subject not visible on GitHub
 - [x] octobox.io
 - [ ] My instance

🐞 **Problem**

An inline comment/reply box appeared on Octobox.io 🎉 
But it doesn't work 😢 

Responses do show up in Octobox, but not on GitHub.
I've tried re-sync / awaiting a little time to see if things would sync up, but it doesn't.

Visible in Octobox:
image

GitHub:
image

🎯 **Goal**

Well, it is an exposed feature, which seems to work, but doesn't.
Which is tricky.

💡 **Possible solutions**

Disable inline issue/pr view.

n/a

BenJam:
fixes #1632 (I hope)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -157,7 +157,6 @@ class User < ApplicationRecord
     return false unless subject.commentable?
     return true if personal_access_token_enabled?
     return true if Octobox.fetch_subject?
-    return true if subject.repository.open_source?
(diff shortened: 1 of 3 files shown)


andrew:
I blame GitHub's complete lack of oauth documentation!

PR #1641, opened 2019-02-26
Update app installation sync rake task

andrew:
Because GitHub doesn't tell you when a user has upgraded the permissions on their app


BenJam:
yay-hub.

PR #1642, opened 2019-02-26
Add syncing icon to comments that are not yet on GitHub

BenJam:
Simple visual indicator that comments are not _exactly_ live. Also helps inform users if there's ever an issue on the server with comment sending.


andrew:
Will this ever stop spinning if you leave the page open?

BenJam:
Refreshes once the GH info is tacked onto the comment. I'll load it onto my instance.

BenJam:
++

PR #1644, opened 2019-02-26
fix JS comment form keydown handler

BenJam:
`s/event/e`


BenJam:
err. why u fail travis?

PR #1645, opened 2019-02-26
Launch 'conversations'

BenJam:
- [x] remove beta banner from user settings panel
- [x] default all new users to open threads in octobox
- [x] run rake task to update all users to open threads in octobox


db/migrate/20190301131405_display_comments_in_ocotobox.rb
```diff
@@ -0,0 +1,13 @@
+class DisplayCommentsInOcotobox < ActiveRecord::Migration[5.2]
+  def up
+  	User.all.find_each do |user|
+  		user.update_attribute(:display_comments, true)
```
andrew: In general we want to avoid migrations that change data using application code, because it ties the migration to a point in time.

Instead the contents of this migration should probably be:

```ruby
def change
  change_column :users, :display_comments, :boolean, default: true
end
```

lib/tasks/tasks.rake
```diff
@@ -48,6 +48,14 @@ namespace :tasks do
     AppInstallation.sync_all
   end
 
+  desc "Update all users to display comments in Octobox"
+  task display_comments: :environment do
+    User.all.each do |user|
```
andrew: You can do this with a single sql update: `User.update_all(display_comments: true)`

andrew:
There's a few bugs with the thread commenting that should get squashed before this gets turned on for everyone.

BenJam:
Right, let's get this over the line and out as the default behaviour. The marketing benefit of having the infinitacle all over GH is too good to miss out IMO.  Looks like #1478 is holding us up and I'm not sure what's happening there...

andrew:
@BenJam I think https://github.com/octobox/octobox/pull/1760 will help with #1478

BenJam:
anything else you can think of that needs to be sorted?

andrew:
Nothing else that I can think of, this can be merged so that all new signups get the new default straight away, then run the rake task along with an announcement blog post.

PR #1648, opened 2019-02-27
Add delay to display of [data-toggle=popover]

### LINKED ISSUE #1647: Issue hidden when moving from checkbox to title
 - [x] octobox.io
 - [ ] My instance

🐞 **Problem**
I regularly use octobox to triage issues by subject, marking the issues I suspect I will archive with a check and then clicking through the link in another tab to double check before coming back to continue down my list. In the last couple months as the Github App has been implemented I keep tripping up on the very eager tooltips on the issue/pr icon that interupts my flow

🎯 **Goal**
I'm unable to install Github Apps for our organization, Octobox is currently in review and hopefully we'll be able to activate the Github App for our org soon but in the meantime I'm feeling quite overwhelmed by the number of reminders to activate the Github App which I can do nothing about. And they're especially hard to ignore when they get caught under my mouse and prevent me from clicking the issue.

💡 **Possible solutions**
I'm thinking a 2 or 3 second delay on showing popovers would be appropriate, similar to the delay browsers use before showing alt text. (I'm thinking this should also be applied to the "Star" label on the obvious star icon). Alternatively, the prompt to install the Github App could be shown as a modal or something on page load, or maybe a header bar of some sort that can be dismissed.

I'd be happy to submit a PR if someone would please let me know what the desired end state is. If it's just a delay then I should be able to implement that no problem here: https://github.com/octobox/octobox/blob/master/app/assets/javascripts/octobox.js

spalger:
Fixes #1647

In order to avoid showing the popovers on individual issues when just quickly moving the mouse over the notification icon this adds a 2 second delay to showing the popover. I looked around for a way to test this but there doesn't seem to be any JS tests that I could find. Additionally, when I setup and ran the project locally I didn't see the popup at all. I'm confident in the changes, but if someone could help me verify them that'd be great.


BenJam:
Thanks for putting this together, I'll take a look asap.

spalger:
@BenJam ping 😃

BenJam:
Ah sorry, I recall taking a look at this a little while back but there was something up with it. Gimme a tick...

BenJam:
Urgh, managed to mess up my local so i’ll take a look when I have that done. Looks okay though.

BenJam:
okay so sorry or delay but I get the following error when the event trips on 2s hover:

```
Uncaught TypeError: this.element.getAttribute is not a function
    at Popover._fixTitle (tooltip.self-07c32e7ffc85c7766b3626f1ebb36fff50d7bbd0db98d67a01957fb594430c7d.js?body=1:662)
    at Popover._setListeners (tooltip.self-07c32e7ffc85c7766b3626f1ebb36fff50d7bbd0db98d67a01957fb594430c7d.js?body=1:657)
    at Popover.Tooltip (tooltip.self-07c32e7ffc85c7766b3626f1ebb36fff50d7bbd0db98d67a01957fb594430c7d.js?body=1:304)
    at new Popover (popover.self-7230e09abae7f369855a6e32aa734934ee3d3d9907dd72548df3e6a1d0284248.js?body=1:130)
    at Popover.toggle (tooltip.self-07c32e7ffc85c7766b3626f1ebb36fff50d7bbd0db98d67a01957fb594430c7d.js?body=1:333)
    at tooltip.self-07c32e7ffc85c7766b3626f1ebb36fff50d7bbd0db98d67a01957fb594430c7d.js?body=1:633
    at dispatch (jquery3.self-5af507e253c37e9c9dcf65064fc3f93795e6e28012780579975a4d709f4074ad.js?body=1:5184)
    at elemData.handle (jquery3.self-5af507e253c37e9c9dcf65064fc3f93795e6e28012780579975a4d709f4074ad.js?body=1:4992)
``` 

Looks like the context isn't being retained. Incidentally I think a 2s hover is quite a long time, maybe 0.5-1s more appropriate?

spalger:
Ah, yes, my bad. That should fix it.

PR #1655, opened 2019-03-01
Added integration tests of error pages

jhsu802701:


jhsu802701:
I now believe that some of these error controllers/pages are not actually used in the app.

PR #1656, opened 2019-03-02
Removed unused error pages

jhsu802701:
A search of the files in this app showed that the '/404' and '/500' pages are not used.  This pull request removes the "not found" and "internal error" routes, pages, and controllers.


pradyunsg:
They aren't directly used. They are used when a page is not found (resulting in a 404 HTTP status) or there's a server-side error (resulting in a 500 HTTP status).

See http://octobox.io/sdgjgjsbgjkbs for a case where the 404 template is used.

jhsu802701:
I don't see the 404 template at that URL.  Instead, I'm redirected to the root page.

andrew:
404 pages only seem to render when logged in, rather than removing the code, we should fix it so it uses it all the time.

Issue #1662, opened 2019-03-04
Comment edits don't show up in thread view


akoeplinger:
Are you experiencing this issue in octobox.io or your own instance?
 - [X] octobox.io
 - [ ] My instance

🐞 **Problem**

When a comment is edited on an issue the thread view in Octobox doesn't show the update, it displays the original comment.

🎯 **Goal**

Edits to comments should be reflected in Octobox.

BenJam:
Yup, it would be great to have this.

BenJam:
👍

PR #1664, opened 2019-03-05
Update comments in thread when updated on the DB.

BenJam:
I noticed that the sync icon persists on a new comment because the comment view isn't updated when the model is updated with a GitHub id. 

Should also be useful for #1662 at some point.

### DIFF
--- a/app/assets/javascripts/channels/comments.js
+++ b/app/assets/javascripts/channels/comments.js
@@ -9,6 +9,8 @@ var subscribeToComments = function(){
     received: function(data){
       if ($('#notification-thread').attr('data-id') == data.subject_id && !$("#comment-"+data.comment_id).length){
         $('.discussion-thread').append(data.comment_html);
+      }else if ($("#comment-"+data.comment_id).length){
+        $("#comment-"+data.comment_id)[0].outerHTML = data.comment_html;
       }
     }
   });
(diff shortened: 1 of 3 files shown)


app/models/comment.rb
```diff
@@ -1,7 +1,7 @@
 class Comment < ApplicationRecord
   belongs_to :subject
 
-  after_create :push_to_channels
+  after_save :push_to_channels
```
andrew: This could potentially create a LOT more background jobs, some of which may be no-ops, we should make sure that we only push to channels if the visible contents of the comment has changed, see https://github.com/octobox/octobox/blob/master/app/models/notification.rb#L190-L196

app/assets/javascripts/channels/comments.js
```diff
@@ -9,6 +9,8 @@ var subscribeToComments = function(){
     received: function(data){
       if ($('#notification-thread').attr('data-id') == data.subject_id && !$("#comment-"+data.comment_id).length){
         $('.discussion-thread').append(data.comment_html);
+      }else if ($("#comment-"+data.comment_id).length){
```
andrew:
```suggestion
      } else if ($("#comment-"+data.comment_id).length){
```

app/models/comment.rb
```diff
@@ -25,6 +25,14 @@ def unread?(notification)
     notification.last_read_at && DateTime.parse(notification.last_read_at) < created_at
   end
 
+  def push_if_changed
+    push_to_channels if (saved_changes.keys & pushable_fields).any?
+  end
+
+  def pushable_fields
+      ['body', 'github_id']
```
andrew: Bit too much whitespace here

BenJam:
squoosh when ready @andrew

PR #1714, opened 2019-04-01
Improve scrolling behaviour on iOS

andrew:
Based on https://css-tricks.com/the-trick-to-viewport-units-on-mobile/

Behaviour before:

before mov

After:

after mov


BenJam:
nice trick dat.

PR #1755, opened 2019-04-16
Account for plan switches in OC task

BenJam:
Accounts for plan switches in OC task, probably needs a proper test block implemented before merging.


andrew:
:shipit:

PR #1760, opened 2019-04-18
Existing subjet comment_count shouldn't be set to nil if remote comment data missing

andrew:
If a certain subject related webhook/api payload doesn't include a comment count, it shouldn't set an existing comment_count back to nil if it was already a number.

Related to https://github.com/octobox/octobox/issues/1478


BenJam:
👍

PR #1782, opened 2019-04-26
Fix Open Collective rake task

BenJam:
- [x] Re-add removal of accounts
- [x] Add tests
- [x] Remove accounts that have lapsed


BenJam:
@andrew just thinking about lapsed accounts: when does this script get run? I suspect I need to add the last-transaction date to `get_sub_names` and make sure that date+ 1 month is added as the `next_billing_date` on the corresponding subscription purchase.

andrew:
It runs every hour I believe

BenJam:
okay I remembered we filter out tx's at the first step, so I think this is good to go now.

andrew:
👍 merge when ready

PR #1783, opened 2019-04-27
Updated Debian-based local installation documentation.

juliquiron:
`rbenv` is not present in some Debian-like distributions and is assumed to be installed in next commands.

I would suggest to add it to the installation list in the `apt-get` command to make it easy to follow the installation instructions.


andrew:
Thanks!

PR #1784, opened 2019-04-27
Add browser pre-rendering for unread notifications.

juliquiron:
The problem
 - When checking new notifications it takes always time to open the GitHub page in the browser. At the end of the day, this is quite of time.

Proposed solution
 - With the assumption that we are going to see in GitHub most of the unread notifications (or all of them), make sense to prerender the page in order to make fast each page loading


SimonCropp:
This is cool. however for PRs that are merged i very rarely visit the github page. and when using dependabot with automerge, most of my notifications are already merged PRs

juliquiron:
Hi @SimonCropp, thanks for your feedback. This definitively makes sense.

I added an exclusion to the merged PR's in the condition. :+1: 

I used the `notification_icon_color` for it as long in the helpers is what is checking the PR status (https://github.com/juliquiron/octobox/blob/master/app/helpers/notifications_helper.rb) but it lets a quite wired code to read :thinking: . If we go further with this PR I would suggest writing a specific helper to just check the status.

SimonCropp:
@juliquiron thanks. even better

BenJam:
Theoretically I'm cool with this, but we might want to also check that the use isn't using 'thread-view' as a lot of that content would be shown in the thread. Check #1645 for updates as we'll roll this out for everyone soon.

app/views/layouts/application.html.erb
```diff
@@ -42,6 +42,13 @@
     <% if Octobox.config.push_notifications && logged_in? %>
       <meta name="push_notifications" content="true">
     <% end %>
+    <% if @notifications.to_a.any? %>
+      <% @notifications.each do |notification| %>
+        <% if notification.unread? and !['closed', 'merged'].include?(notification.state) %>
```
andrew: Would be great to get this added as a method on the Notification class, something like:

```ruby
def prerender?
  unread? and !['closed', 'merged'].include?(state)
end
```

juliquiron:
Thanks @BenJam and @andrew for your feedback. I moved the `prerender` check to the model and also added to it check if this will be displayed as thread. I guess now is fine.

andrew:
LGTM, thanks for the contribution. Tests are currently broken on master and need to be fixed before this can be merged.

juliquiron:
Branch updated. Now it looks ready to merge :)

andrew:
:shipit:

SimonCropp:
awesome work peoples :)

PR #1812, opened 2019-05-08
Add Timecop and freeze time for open collective tests

BenJam:
Insert gif of JCVD doing the splits here...

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,7 +33,6 @@ gem 'jwt'
 gem 'oj'
 gem 'yard', require: false
 gem 'commonmarker'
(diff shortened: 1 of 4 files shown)


andrew:
```ruby
teardown do
  Timecop.return
end
```

andrew:
Turns out setting a constant of a date in `PERIOD = 1.month.ago` means that never changes, even if Timecop updates it, because it's already been saved, which would also affect the code running in production, in effect not changing the date until the rails app is restarted, so I removed that constant.

BenJam:
👍

PR #1828, opened 2019-05-14
Fix broken UI issue for long words

### LINKED ISSUE #1799: Broken UI for Google Chrome
 - [x] octobox.io

🐞 **Problem**

First of all thanks a lot for this amazing tool, I've been using it for a while and I love it so far. The problem I'm having is more related to the UI, so if some issue / PR title has more than a certian number of characters then the UI looks broken to me and I can't even scroll to it either. For example:

Screenshot 2019-05-03 16 12 30

🎯 **Goal**

To have a consistent UI, that includes the type of the issue, commenter/creator and time. I assume this is happening for the title, so maybe truncate the title to a certain length?

Thanks again, and please let me know if this is desirable then I can probably submit a pull request.

abunashir:
Currently, if a notification's subject is a single word and more than certain characters then it breaks the listing and doesn't show all of the fields

This commit changes a style rule, so longer words will work like any other title, break it down to a certain limit.

Fixes #1799

### DIFF
--- a/app/assets/stylesheets/_notifications.scss
+++ b/app/assets/stylesheets/_notifications.scss
@@ -231,6 +231,7 @@ $notification-items: (
     .notification-subject {
       display: inline-block;
       .link {
+        word-break: break-all;
         overflow-wrap: break-word;
       }
       &:first-child .badge{


andrew:
Thanks for the contribution!

PR #1843, opened 2019-05-22
Add next billing date to OC subscriptions

BenJam:
At the moment users without a GH username on their OC account will be stripped of their subscription when the open collective members are synced. This change adds a next_billing_date to subscription purchases that use the synchronous OC flow to ensure that they're only reevaluated after one month.  

* Simplify logic for removing subscriptions in OC rake task
* Add next_billing_date to open collective synchronous transactions


andrew:
:shipit:

PR #1847, opened 2019-05-23
Don't strip quotes from seach query when parsing.

### LINKED ISSUE #1846: Saved label searches have quotes stripped out
 - [x] octobox.io
 - [ ] My instance

🐞 **Problem**
When I do a search for a label like ":apm" or ":Logs UI", the search initially works and the filter above the search results shows `Search: inbox:true label:":Logs UI"`, but the search box itself will have stripped out the quotes, so if you refresh, you get no results and the filter now shows `Search: inbox:true label::Logs UI` which I believe is an invalid query.

This same problem happens when you save the search, which means saved searches like these will never return results. 

Screen Shot 2019-05-23 at 9 16 21 AM
Screen Shot 2019-05-23 at 9 16 40 AM

🎯 **Goal**
Anyone who uses labels with special characters in them (spaces, colons, etc) can only save searches for those labels if this is fixed.

💡 **Possible solutions**
I imagine the code just needs to stop stripping quotes out of the search string.

If someone points me to where this code lives, I'd be happy to take a stab at fixing it. My Ruby is a bit rusty but I can possibly figure it out if someone can push me in the right direction :)

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

andrew:
Fixes #1846

### DIFF
--- a/lib/search_parser.rb
+++ b/lib/search_parser.rb
@@ -15,7 +15,6 @@ class SearchParser
     while (m = OPERATOR_EXPRESSION.match(query, offset))
       key = m[1].downcase.to_sym
       value = m[2]
-      value = value[1, value.length - 2] if ["'", '"'].include?(value[0])
       offset = m.end(2)
       @operators[key] ||= []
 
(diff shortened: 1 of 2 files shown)


andrew:
This may have broken some other searches 🙈 going to do some further testing now.

Issue #836, opened 2018-08-23
Add filter to distinguish between direct review requests and team review requests


lukashlavacka:
Similar as there is filter for `Mention` and `Team mention` there might be a way to add filter for `Review requested` only for direct review requests and `Team review requested` for requests of a team you are part of.

🐞 **Problem**

Being able to differentiate between `Review requested` and `Team review requested would help especially with reviews requested by `CODEOWNERS` feature vs reviews requested directly.

ahayworth:
Big +1 to this - I would greatly appreciate being able to distinguish when *I* personally must give a review, or when my team in general needs to give a review (it affects how I prioritize my work).

I think the v3 REST API response is enough to distinguish this - the [single PR endpoint](https://developer.github.com/v3/pulls/#get-a-single-pull-request) has both `requested_reviewers` and `requested_teams`.

chrisarcand:
I'm tired of this being a thing, as it's plagued my workflow for literally years now. I was hoping GitHub would just distinguish between `review_requested` and `team_review_requested` in their [list of Notification reasons](https://developer.github.com/v3/activity/notifications/#notification-reasons), but it's been 2.5 years since review requests shipped and by this point I imagine it could be a breaking change anyway (and there's still no V4 Notifications API, I note...) 😕 

I sat down and came up with a potential solution that I'll be alpha testing next week, and will submit it if all goes well. Some early screenshots:

![image](https://user-images.githubusercontent.com/2430490/58373339-5e40c800-7ef2-11e9-9e0a-4dc4e27f2487.png)
![image](https://user-images.githubusercontent.com/2430490/58373344-79133c80-7ef2-11e9-8d89-6f8e2ba2bba0.png)

NicholasEllul:
@chrisarcand any update on the progress of this?

pradyunsg:
See https://github.com/octobox/octobox/pull/1940

chrisarcand:
Coincidentally, there is! A big one: https://github.com/octobox/octobox/pull/2028

banarun:
Are there any plans to schedule #2028?

PR #1859, opened 2019-05-28
Add pull request reviews, pull request review comments and pull request review comments (yes, that's right)

BenJam:
As above. WIP


BenJam:
uses https://github.com/octobox/octobox/pull/1141 in fixtures to test (so check out that thread for expected behaviour). Does not support commit-level comments.

BenJam:
hmmmm, this is currently storing a lot of 'no description given' comments where there are body-less reviews/comments

BenJam:
test currently seems to be skipping `update_comments` in `self.sync`. Any pointers appreciated if anyone gets a look see.

BenJam:
back at this: looks like this isn't calling download comments because it's not getting a gh client.

andrew:
Comment syncing is still optional and enabled with `INCLUDE_COMMENTS=1` ENV var, so it'll need to be turned on in the tests that should sync comments, at least `test/models/subject_test.rb` and `test/controllers/hooks_controller_test.rb`, with the following code:

```ruby
setup do
  stub_include_comments
end
```

test/models/subject_test.rb
```diff
@@ -228,6 +228,26 @@ class SubjectTest < ActiveSupport::TestCase
...
+    reviews = { status: 200, body: file_fixture('subject_58_reviews.json'), headers: { 'Content-Type' => 'application/json' } }
+    comments_for_review = { status: 200, body: file_fixture('subject_58_comments_for_review.json'), headers: { 'Content-Type' => 'application/json' } }
+    review_comments = { status: 200, body: file_fixture('subject_58_review_comments.json'), headers: { 'Content-Type' => 'application/json' } }
+
+    stub_request(:get, remote_subject["url"]+'/reviews').and_return(reviews)
+    stub_request(:get, remote_subject["url"]+'/comments').and_return(review_comments)
+    stub_request(:get, remote_review["pull_request_url"]+'/reviews/'+remote_review["id"].to_s).and_return(comments_for_review)
+
+    Octobox.expects(:include_comments?).returns(true)
+    Subject.any_instance.expects(:update_comments)
```
andrew: This line is actually stubbing out the `update_comments` call, so it doesn't actually run, only asserting that it get's called.

Recommend you remove those two lines and use `stub_include_comments` instead.

BenJam:
ta!

app/models/subject.rb
```diff
@@ -78,7 +78,7 @@ def self.sync(remote_subject)
     return unless subject.persisted?
 
     subject.update_labels(remote_subject['labels']) if remote_subject['labels'].present?
-    subject.update_comments if Octobox.include_comments? && subject.has_comments?
+    subject.update_comments if Octobox.include_comments? && (subject.has_comments? | subject.pull_request?)
```
andrew:
```suggestion
    subject.update_comments if Octobox.include_comments? && (subject.has_comments? || subject.pull_request?)
```

app/controllers/hooks_controller.rb
```diff
@@ -8,7 +8,7 @@ def create
     when 'issues', 'issue_comment'
       SyncSubjectWorker.perform_async_if_configured(payload['issue'])
     when 'pull_request', 'pull_request_review', 'pull_request_review_comment'
-      SyncSubjectWorker.perform_async_if_configured(payload['pull_request'])
```
andrew: I think it'd be good tosplit `'pull_request'` out from the review and review_comment section, so it still calls `SyncSubjectWorker` here

BenJam: will do now

BenJam: I've pushed this but: it'll still sync the comments for the PR as a whole. I could split it out to check for only reviews but I'm not sure there's that much additional work to call the comments on the PR too... Up to you if you want it further split. I'm easy either way.

chrisarcand:
Whoa neat. @BenJam @andrew I don't _think_ anything here conflicts with it, but just as an FYI to keep in mind with this review request area, here's the branch I'm currently testing for #836, if we want to ensure this all lines up nicely : https://github.com/octobox/octobox/compare/master...chrisarcand:add-team-vs-individual-review-requests

andrew:
@chrisarcand that's looking great, this PR only pulls in the review related comments, doesn't touch any requests. That branch is looking great, can't wait for the PR!

andrew:
@BenJam do the review comments visually look any different to regular comments?

BenJam:
it's all one comment stream atm.

andrew:
@BenJam perhaps we can add store the `state` field if it's present (which it only will be for reviews) and then add an icon to the header of the comment.

app/models/subject.rb
```diff
@@ -203,8 +204,10 @@ def download_reviews
     return unless github_client && pull_request?
     reviews = github_client.get(url + '/reviews', since: comments.order('created_at ASC').last.try(:created_at))
     reviews.each do |review|
-        reviews.concat download_comments_for_review(review) if review[:body].empty?
+      if review[:state] == "COMMENTED"
+        reviews.concat download_comments_for_review(review)
```
andrew: I believe that users can leave comments when the review is in any state, not just commented

BenJam: Yeah that’s a mental typo.

BenJam:
@chrisarcand I can't see anything that's going to screw up your PR significantly by merging this. I'd quite like to get it out and move everyone onto comment-threads asap.

chrisarcand:
For sure, go for it, looks great!

BenJam:
Cool, just going to address a little bug in which reviews aren't stored with the correct datetime

BenJam:
Such meta:

Screenshot 2019-06-18 at 13 51 12

bertie:
Hey!

PR #1896, opened 2019-06-18
Updated db/schema.rb

jhsu802701:
This is just the update from running the routine database migration, which is not reflected in the current code base.


jhsu802701:
I have synced this branch.

andrew:
Thanks!

PR #1915, opened 2019-07-05
Update application icons

BenJam:
Standardises octobox icons in /public so the icon looks good when saved to the homepage of android/iOS devices: 

IMG_6852


andrew:
@benjam related issue: https://github.com/octobox/octobox/issues/1916

PR #1937, opened 2019-07-16
Only download subjects for a notification if the user is present

andrew:
Avoids an edge case where the user has deleted their account whilst their notifications are being synced


BenJam:
LGTM

PR #1939, opened 2019-07-16
Swallow internal server errors from Octokit when syncing review comments

BenJam:
Pockets internal server warnings in order to stave off a balloon in logging.

### DIFF
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -199,7 +199,7 @@ class Subject < ApplicationRecord
   def download_comments
     return [] unless github_client
     github_client.get(url.gsub('/pulls/', '/issues/') + '/comments', since: comments.order('created_at ASC').last.try(:created_at))
-  rescue Octokit::ClientError => e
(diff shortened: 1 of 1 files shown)


app/models/subject.rb
```diff
@@ -199,7 +199,7 @@ def download_status
   def download_comments
     return [] unless github_client
     github_client.get(url.gsub('/pulls/', '/issues/') + '/comments', since: comments.order('created_at ASC').last.try(:created_at))
-  rescue Octokit::ClientError => e
+  rescue Octokit::ClientError, Octokit::InternalServerError => e
```
andrew:
```suggestion
  rescue Octokit::ClientError => e
```

app/models/subject.rb
```diff
@@ -214,23 +214,23 @@ def download_reviews
       end
     }
     return reviews
-  rescue Octokit::ClientError => e
+  rescue Octokit::ClientError, Octokit::InternalServerError => e
```
andrew:
```suggestion
  rescue Octokit::ClientError => e
```

PR #1803, opened 2019-05-06
Upgrade rails to 6.0.0

andrew:
Preparing for rails 6 by testing out 6.0.0.rc1

PR #1940, opened 2019-07-16
[POC]: Add team vs individual review requests

chrisarcand:
_**This is a proof of concept/work in progress. Don't merge it yet.**_

_Opening this for others to try if they like, as well as to open the door for UI help as I think it needs a little help there (I like the person vs multi person icons, but they look misaligned last I saw)._

---

The GitHub Notifications API doesn't distinguish between individual and team review requests for a reason field.  Regardless if its for a team you're on or you specifically, you get the same reason and there's no distinction.

...that sucks, a lot. A lot of teams work where someone belongs on a team that gets pinged based on the code owners functionality, but it only means *someone* on the team should review it.  Therefore, 'review requested' gets ignored in the huge number of team requests that look identical to when someone really wants the individual specifically to review it. Whoops.

It's been nearly three years since review requests were added to GitHub, and we've been holding out that they would actually distinguish it eventually. But I thought about it more and realized it's probably considered a breaking change and by now it's never going to happen.

So, I cave! This change introduces the first time we've ever intentionally set a notification reason that isn't just pulled directly from the GitHub API, but it's an extremely useful one that I absolutely need.

On subject syncs, query for all notifications that would affect someone in Octobox who was review requested and set team review requests based on the current state of the subject, which contains lists of reviews by individual and team.

See individual commit messages for additional details.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -301,9 +301,7 @@ class NotificationsController < ApplicationController
   def load_and_count_notifications(scope = notifications_for_presentation.newest)
     @types                 = scope.reorder(nil).distinct.group(:subject_type).count
     @unread_notifications  = scope.reorder(nil).distinct.group(:unread).count
-    @reasons               = scope.reorder(nil).distinct.group(:reason).count
     @unread_repositories   = scope.reorder(nil).distinct.group(:repository_full_name).count
-
     @states                = scope.reorder(nil).distinct.joins(:subject).group('subjects.state').count
     @statuses              = scope.reorder(nil).distinct.joins(:subject).group('subjects.status').count
     @unlabelled            = scope.reorder(nil).unlabelled.count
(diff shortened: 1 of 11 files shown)


BenJam:
I’ll take a look at this in the morning but first and foremost: thanks! 🤙

andrew:
Had a browse through the code, looks like a reasonable way to add in the new reason type 👍 

I've yet to check it out and have a play with it, although I'm not sure I have any team review requests to sync in. 

One thing I'm wondering is the likelihood of the most recent reason changing from review request and if that means some users may fall through the cracks where their reason keeps changing, but we're not exactly in control of that, I'd hope that the reason logic on GitHub's side ends up preferring to stick with the more important reasons (like reviews).

andrew:
I've also requested a review from @octobox/core here to get it to show up when I test locally.

BenJam:
@andrew Yeah, with a conversation also occurring in the PR, the reason could easily flip to mention/commented etc. (which is what I think has happened to me already... and now to you :P)

In principle this looks cool @chrisarcand 

I do get a fail on `test/models/subject_test.rb:202` and  `test/models/subject_test.rb:208` but i assume that's expected because TDD?

I also get a deprecation warning re. rails 6.0 
```
reason = 'team_review_requested' DESC, reason ASC". Non-attribute arguments will be disallowed in Rails 6.0. This method should not be called with user-provided values, such as request parameters or model attributes
```
which we will probably want to avoid re. #1803.

chrisarcand:
Yeah, because this feature was written two months ago and I haven't touched it since, expect tests to maybe fail and this project definitely wasn't on the Rails 6 RC yet. 

After using it a day or two I've noticed something is definitely not quite right, I'm getting Review Requested (individual) for some team reviews. 😑 Digging will continue!

BenJam:
No worries! We're happy to guide where needed and equally happy to leave you to it. This feature would be cool though so let us know what we can do if anything.

chrisarcand:
Closing in favor of #2028 

This turned out to not work well because there are just way too many edge cases and moving parts in the awkward state machine that you're reverse engineering with reasons in the GitHub API.

PR #1944, opened 2019-07-19
wrap threads in card so that sticky header stays in place on mobile

BenJam:
before:
before

after:
cap


BenJam:
there might be a bit of weridness going on with the gif encoding there with that white line, I can't see it on my actual device.

andrew:
🚢

PR #1724, opened 2019-04-06
Store draft boolean on subjects

andrew:
Closes https://github.com/octobox/octobox/issues/1613

Todo:

- [x] grey icon for open draft pull requests
- [x] draft link in sidebar
- [x] draft search prefix
- [x] draft displayed in threadview
- [x] tests

Screenshot 2019-04-06 at 21 26 58
Screenshot 2019-04-06 at 21 26 55
Screenshot 2019-04-06 at 21 26 50

PR #1945, opened 2019-07-19
Add draft state to the API

lugray:
Follow-up to #1724 to surface the draft state in the API.  As it stands, you can filter in the api with notifications.json?draft=true, but not determine this information when unfiltered.


andrew:
🚢

PR #1985, opened 2019-08-22
Pinned searches count

jules2689:
Introduces a new pinned_search#show.json endpoint for internal use that includes the count of the search, this then leverages the endpoint by asynchronously populating the count badge for each saved search.

I chose to do this asynchronously to anticipate that one user with like 40 saved queries that would time out if we did this normally.

If the async request fails, we just remove the count badge. Until load we show a small syncing icon spinning to indicate a request is going on.

Slowed down gif for reference:
octobox_counts

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -359,19 +359,16 @@ class NotificationsController < ApplicationController
   end
 
   def notifications_for_presentation
-    eager_load_relation = [{subject: :labels}, {repository: {app_installation: {subscription_purchase: :subscription_plan}}}]
-    scope = current_user.notifications.includes(eager_load_relation)
-
-    @search = Search.new(scope: scope, query: params[:q], params: params)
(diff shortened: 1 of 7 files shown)


app/controllers/notifications_controller.rb
```diff
@@ -359,19 +359,16 @@ def current_notifications(scope = notifications_for_presentation)
   end
 
   def notifications_for_presentation
-    eager_load_relation = [{subject: :labels}, {repository: {app_installation: {subscription_purchase: :subscription_plan}}}]
-    scope = current_user.notifications.includes(eager_load_relation)
-
-    @search = Search.new(scope: scope, query: params[:q], params: params)
+    @search = Search.initialize_for_saved_search(query: params[:q], user: current_user, params: params)
```
jules2689: I re-used this scope to make a more efficient query, so I abstracted to a new constructor on search :)

PR #2013, opened 2019-09-08
fix: Add ALLOW_HOSTNAME and ALLOW_ALL_HOSTNAMES environment variables

### LINKED ISSUE #2011: Allow addding hostname via environment variable
 - [ ] octobox.io
 - [X] My instance
 

🐞 **Problem**

When running the latest version of octobox on OpenShift I run into this error:

grafik

The reason is this new Rails feature: https://www.fngtps.com/2019/rails6-blocked-host/

I.e. one would have to add either the hostname of the reverse proxy to this list or enable a full whitelist. Ideally, this could be done via an environment variable (e.g. `OCTOBOX_HOSTNAME` or `OCTOBOX_NO_HOSTAUTHORIZATION`)as there is no easy way to edit this configuration file when installing octobox as it is part of the Docker image.

An alternate solution would be to adapt the OpenShift installation to include a local slim reverse proxy which calls rails on localhost. This might be the more generic solution.

rhuss:
This is to allow an octobox setup in an reverse proxy setup and to pass
the hostname check as described in https://www.fngtps.com/2019/rails6-blocked-host/

Also, adapted the OpenShift templates to use ALLOW_ALL_HOSTNAMES as
octobox is shielded by the OpenShift router anyway and does not require
this hostname check.

Fixes #2011.


rhuss:
Sorry, but I have no idea how I should test this change. Tbh, I'm not a Ruby guy and although I probably could adapt an existing test case, for setting up a new test I would have to invest more time than I am currently able to spend.

'happy though to handover this PR to someone with more testing experience, as I think this is an important feature required for any setup with Octobox running behind a reverse proxy (like within a Kubernetes cluster with Ingress or in OpenShift, proxied through the OpenShift router).

The other change to the OpenShift template is also necessary to update to Redis 4 (which is the minimum Redis version now supported, was 3 before the latest changs).

sorry ...

acidicX:
@andrew any news on this? The missing env is blocking Docker updates.

andrew:
@acidicX I've been away on holiday, looking at getting this merged soon

acidicX:
@andrew :raised_hands: thank you! Hope you had a nice holiday :)

PR #2037, opened 2019-09-25
add author to json response for notifications#index

gabrielpoca:
Adds the `author` field to the notification's subject in the json
response when that same field would have been available in the html
response.

This is useful for API clients because there is no other way to get the
author for a notification.


gabrielpoca:
I hope everything is good with the PR. From reading the source code I'm assuming this change doesn't require a test.

gabrielpoca:
hey @andrew I hope you don't mind me tagging you. I noticed you approved and merged many of the last commits, I would like to ask if this contribution is welcome or if I should instead just maintain my fork?

andrew:
@gabrielpoca yeah this looks good, can you add a test for it and I'll get it merged.

gabrielpoca:
@andrew so, I found the tests, at first I was looking for view tests, but I see now that they are in the controller. Added, I hope this is is good, it's has been a couple of years since I last programmed in ruby

PR #2108, opened 2019-11-22
Drop support for Mysql

andrew:
In an effort to reduce the long term overhead of maintaining octobox I'd like to drop support for mysql as an alternative database, I don't believe that many people are using that option and it only causes headaches when wanting to use more advanced features in postgres like array and json fields.

If you or your company is using octobox with mysql and would like to see it continue to be supported, please consider supporting the project financially: https://opencollective.com/octobox/


BenJam:
👍

PR #2138, opened 2019-12-12
Fix link to API Documentation page under main Documentation page.

### LINKED ISSUE #1593: Link to Octobox API
When I view the documentation page of the local app in my web browser, the link to "Octobox API" (under "API Documentation") leads to "/docs", which leads to the "No route matches" error.  On the Octobox.io site itself, this isn't a problem, as this actually leads to the API documentation page, which seems to be outside the scope of this Rails app.

Should this OctoBox API link lead to https://octobox.io/docs?

ghosteathuman:
Fixes #1593


andrew:
Thanks!

PR #2140, opened 2019-12-13
Fix spinner animation when background jobs is not enabled.

### LINKED ISSUE #913: No refresh spinner anymore
 - [ ] octobox.io
 - [X] My instance

🐞 **Problem**

With the latest update (I think already since some days), refreshing the notification list with either the top-level button or the central button on an empty list doesn't provide any feedback anymore that it's refreshing. Previously there was a spinner during refresh.

This resembles a bit https://github.com/octobox/octobox/issues/829#issuecomment-414927819 but the key shortcut 'r' still works.

It still just looks like

image

during refresh, without any animation.

ghosteathuman:
Fixes #913 

Sync spinner will now animate correctly even when `OCTOBOX_BACKGROUND_JOBS_ENABLED=false`

### DIFF
--- a/app/assets/javascripts/octobox.js
+++ b/app/assets/javascripts/octobox.js
@@ -257,7 +257,11 @@ var Octobox = (function() {
     if($("a.js-sync.js-async").length) {
       $.get("/notifications/sync.json", refreshOnSync);
     } else {
+      if(!$(".js-sync .octicon").hasClass("spinning")){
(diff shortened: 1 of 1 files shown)


ghosteathuman:
Sync spinner will now animate correctly even when `OCTOBOX_BACKGROUND_JOBS_ENABLED=false`

ghosteathuman:
@andrew I have rebase against master branch so it ought to be up to date.

PR #2312, opened 2020-04-23
Only sync subjects for users who've been active within the past month

andrew:
Reduce the amount of background jobs being queued for inactive users, save on server resources plus faster syncing for active users 👍


BenJam:
👍

PR #2433, opened 2020-09-09
docs(install): removes left-over bracket

Addono:
Nothing big, just a small (syntax) typo adding a redundant `]` after a link.


andrew:
Good spot, thanks for opening the PR!

PR #2611, opened 2021-03-09
hacky way to disable confirming mute

nornagon:
I work on a very noisy repository (electron/electron), and muting threads is something I do a lot. The extra `confirm()` is quite irritating 😅 

I'd love to have a setting in the preferences for this, but this is an easy way and would work well enough for me!

### DIFF
--- a/app/assets/javascripts/octobox.js
+++ b/app/assets/javascripts/octobox.js
@@ -159,7 +159,7 @@ var Octobox = (function() {
   };
 
   var mute = function(ids){
-    var result = confirm("Are you sure you want to mute?");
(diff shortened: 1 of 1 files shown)


andrew:
@nornagon I think I can add a proper preference this week if you can wait a couple days?

nornagon:
@andrew sure :) no hurry.

andrew:
Just deploying the new setting now, you'll find on this page: https://octobox.io/settings

Screenshot 2021-03-14 at 16 47 14

nornagon:
thanks!

PR #2677, opened 2021-05-18
bold unread message subjects

corywatilo:
I've found the unread notification state a little hard to differentiate from ones I've already read. This helps a little by bolding the unread title, similar to how many email clients handle it.

image


andrew (approved):

andrew:
Thanks!

PR #2852, opened 2021-11-18
Seperate API namespace and notifications controller

### LINKED ISSUE #2817: A custom per_page option in the API shouldn't be stored in a cookie
A weird interaction I ran into recently:

- made a JSON API Request to https://octobox.io/notifications.json?per_page=1 in the browser
- then load the default notifications list in https://octobox.io
- which redirects to https://octobox.io/?per_page=1 

This is because whenever the per_page attribute is changed, it's recorded in a cookie and reused, code here: https://github.com/octobox/octobox/blob/master/app/controllers/notifications_controller.rb#L455

In the short term, the cookie lines should be disabled if the request type if json, in the longer term, the API responses should probably be pulled out into their own separate controller to avoid this kind of bug.

### LINKED ISSUE #2849: Move the api to it's own controller under /api
Currently the json api logic is mixed in with the regular http/html endpoints, whilst it reduces code, as I look at extending the api it's becoming messy and complicated.

Instead all of the api functions should be moved to /api and separate Api controllers.

For backwards compat, we'll leave the existing functionality in place in the notifications controller but mark as deprecated and look at removing that later down the line.

andrew:
Work in progress

Notes for myself:

- [x] shouldn't load sidebar counts 
- [x] shouldn't work for cookies 
- [x] pagination shouldn't remember cookies (fixes #2817)
- [x] update documentation
- [x] move users controller api endpoints
- [x] move pinned_searches controller api endpoints

In future:
- [ ] support for threads (https://github.com/octobox/octobox/issues/2855)
- [ ] api requests to regular notifications controller should be redirected (https://github.com/octobox/octobox/issues/2856)

Closes #2849


andrew:
The new api is now live on octobox.io, docs here: https://octobox.io/docs/

PR #2871, opened 2021-11-30
Fix support for complex labels in search parser

### LINKED ISSUE #2870: Search parsing breaks on certain characters
 - [x] octobox.io
 - [ ] My instance
 
🐞 **Problem**

If a label contains a colon (`:`) or a hyphen (`-`), the search parser transforms the label, making Octobox return the wrong results. I can only verify the problem with these two characters, but there might be other cases as well.

Example queries:
1. 
My query: `label:"kind: bug"`
Transformed query: `label:" kind:bug"`

2.
My query: `label:"created-by: Next.js team"`
Transformed query: `label:"created -by:Next.js team"`

🎯 **Goal**

Defining pinned searches show 0 results for most of my use cases. Typing these manually into the search bar does not work either. The returned result should match the exact query given by the user.

💡 **Possible solutions**

The problem seems to come from `lib/search_parser.rb`, from this change:

https://github.com/octobox/octobox/commit/d76db3d6a92dd8102a1b25bf6c82386b5d82ca75#diff-44b2718766fa20dc1d7738a61afb283b787bcb352a378de349ace75e8a55c455R11

I recently started using Octobox, but one of my colleagues reported the issue being present since that commit has landed.

The Regular Expression that parses the search query should take into account that certain characters can also appear as part of the label (and other groups) terms.

I am currently looking into this myself, but as I have no experience with Ruby local development, it might take longer for me to figure out how to set it up than for the author to find a fix. Unless it's confirmed as being worked on by someone else, I'll continue trying to create a fix, and if I manage to, I'll open a PR.

andrew:
Closes #2870

### DIFF
--- a/test/lib/search_parser_test.rb
+++ b/test/lib/search_parser_test.rb
@@ -73,14 +73,14 @@ class SearchParserTest < ActiveSupport::TestCase
   end
 
   test 'has multiple operators with \' and "' do
-    query = 'operator: \'123 is " lol\' otheroperator: \'12\' operator: "other \' value"'
+    query = 'operator: \'123 is lol\' otheroperator: \'12\' operator: "other value"'
     search = SearchParser.new query
-    assert_equal search[:operator], ["123 is \" lol", "other ' value"]
+    assert_equal search[:operator], ["123 is lol", "other value"]
   end
 
   test 'has multiple operators with \' and " return as array' do
-    query = 'operator: \'123 is " lol\' otheroperator: \'12\' operator: "other \' value"'
-    values = ['123 is " lol', 'other \' value']
+    query = 'operator: \'123 is lol\' otheroperator: \'12\' operator: "other value"'
+    values = ['123 is lol', 'other value']
     search = SearchParser.new query
     assert_equal search[:operator], values
   end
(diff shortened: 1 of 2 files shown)


test/lib/search_parser_test.rb
```diff
@@ -96,4 +96,16 @@ class SearchParserTest < ActiveSupport::TestCase
...
+  test 'allows : within quotes' do
+    query = 'label: "kind: bug"'
+    search = SearchParser.new query
+    assert_equal ['kind: bug'], search[:'label']
+  end
+
+  test 'allows - within quotes' do
+    query = 'label:"created-by:Next.js team"'
+    search = SearchParser.new query
+    assert_equal ['created-by:Next.js team'], search[:'label']
```
balazsorban44, line 109:
```suggestion
    assert_equal ['created-by: Next.js team'], search[:'label']
```

andrew:
Leaving myself a note here to commit later, going to rewrite the parser to use `StringScanner`:

```ruby
@operators = {}

ss = StringScanner.new(" foo:bar -bar: baz baz: 'blah' label:'next: js' label:\"diff:hard\" freeeeeeeetext dsfs")

# scan for key
ss.scan_until(/[^'"](\-?\w+):\s?/)
key = ss.captures[0].to_sym
@operators[key] ||= []

# scan for value
value = ss.scan(/(("|')[^"']+('|")|(\+))|\w+/) 
value.split(',').each{ |v| @operators[key] << v }

# repeat until nil

# freetext 
@freetext = ss.rest 
```

andrew:
Almost ready, one failing test left

PR #3111, opened 2022-08-26
fix: issue #2867

### LINKED ISSUE #2867: Toggling stars results in incorrect star icon until page refreshes
star

KarlHeitmann:
Fix #2867 

Toggling stars results in incorrect star icon until page refreshes
fixed. It was missing the correct inner <path> tag and toggling another
class. Problem is due because the `<%= octicon ... %>` helper is being
used in the .erb view, and this helper is not available in jQuery.

image

Here you an see on file `app/views/notifications/_notification.html.erb` the `octicon` helper receiving as input `star-active` or `star-inactive` depending whether the notification is starred or not. The same class used as input in the javascript `app/assets/javascripts/octobox.js` the difference is that the helper `octicon` is not only toggling the class `star-active <> star-inactive`, it is doing a lot more!

I would like to retrieve the correct `path`using the octicon helper inside `app/assets/javascripts/octobox.js`, but I'm affraid the helper is out of scope from the javascript. One way to solve it, would be using `ujs` instead of plain jQuery. Using ujs, one can embed ruby code inside the javascript, thus leverage `octicon` helper inside the javascript.

Any feedback to improve this PR would be appreciated.

### DIFF
--- a/app/assets/javascripts/octobox.js
+++ b/app/assets/javascripts/octobox.js
@@ -336,12 +336,21 @@ var Octobox = (function() {
 
   var star = function(id){
     $('#notification-thread').data('id') == id ? $('#thread').find('.toggle-star').toggleClass("star-active star-inactive") : null;
-    $("#notification-"+id).find(".toggle-star").toggleClass("star-active star-inactive");
+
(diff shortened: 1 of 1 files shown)


andrew:
Looks good, thanks!

PR #3288, opened 2023-03-23
Use ruby 3.2.1

heka1024:
Ruby 3.2.1 is released, so use it instead of 3.2.0

https://www.ruby-lang.org/en/news/2023/02/08/ruby-3-2-1-released/


andrew:
Thanks!

PR #3302, opened 2023-04-08
Make selection borders more visible

### LINKED ISSUE #3287: make item selection more visually clear
## 💥 Problem

When an item is selected by hovering the mouse over it, the selection is visually reasonably clear, due to the slightly darker background colour:

image

However, if the selection is made via `j` / `k` keyboard shortcuts, there is no change in background colour for the selected item, and the only visual hint of that selection is an extremely thin and hard to spot vertical line:

image

The difficulty of spotting this vertical line is compounded by the fact that if a different item has the mouse over it (as seen here with the item two lines below), that item does not revert back to the default appearance even though it is no longer selected.

I have very good eyesight (when using my glasses or contact lenses!), and even so as someone who prefers to use keyboard over mouse in general, I find it overly difficult to spot this thin vertical line when selecting notifications via keyboard.  It must be a lot harder for someone who is unable to achieve good eyesight.

## Goal
Make the selection visually much clearer.

## Proposed solution
Increasing the width of this vertical line, and perhaps choosing a colour for it which has a much higher contrast to the other colours (i.e. not a grayscale colour) seems like an incredibly quick and easy fix to this problem.

It might also be worth deactivating the mouse hover selection indicators when the keyboard selection deviates from it.

aspiers:
**PLEASE NOTE: this is untested - just a quick hack based on experiments with the userstyle listed in #3287.**

As per #3287, it's really hard to spot which item is currently selected or focused (e.g. via hover) for potential future selection.

So increase the border width from 1px to 5px.

Fixes #3287.

### DIFF
--- a/app/assets/stylesheets/_notifications.scss
+++ b/app/assets/stylesheets/_notifications.scss
@@ -77,19 +77,19 @@ $notification-items: (
 
 @mixin add-borders{
   .table-notifications tr td:first-child {
-    border-left: 1px solid $list-group-bg;
+    border-left: 5px solid $list-group-bg;
   }
   .table-notifications tr.active td:first-child {
-    border-left: 1px solid $list-group-action-active-bg;
+    border-left: 5px solid $list-group-action-active-bg;
   }
   .table-notifications tr:hover td:first-child {
-    border-left: 1px solid $list-group-action-hover-color;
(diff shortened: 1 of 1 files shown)


andrew:
Thanks!

PR #3633, opened 2024-03-02
Add openapi.yaml to the repository.

rickwysocki:
I created an openapi.yaml spec file for Octobox. I relied both on [existing documentation](https://octobox.io/docs/) and my own testing of the endpoints in a Docker version of the application. I have a points/questions to review:

- I don't think I was able to get the sync/syncing endpoints, in the notifications controller, to work in my testing. Sync (/api/notifications/sync.json) might have worked since, based on the existing documentation, there wouldn't be a response, but I couldn't find any evidence of it working. Syncing (/api/notifications/syncing.json) only ever returned full HTML pages in Terminal. I may have misinterpreted elements of the existing documentation here.
- I couldn't seem to post or patch pinned searches. Both returned HTML pages in Terminal. Additionally, the response examples in the existing documentation confused me, so I don't think the responses are currently correct. I'd love to work more on this if you have any thoughts.
- I've left a few other in-text comments for smaller points, such as info I wasn't sure of or didn't have.

This was a lot of fun! I'd love to keep working on this with your feedback on the current version. 

EDIT: One last point I forgot to mention: I was unsure where the openapi.yaml file should go, so it is in the root directory.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -10,7 +10,7 @@ info:
     # email: ""
   license:
     name: "GNU AFFERO GENERAL PUBLIC LICENSE"
-    url: "https://github.com/octobox/octobox?tab=AGPL-3.0-1-ov-file#readme"
+    url: "https://github.com/octobox/octobox/blob/master/LICENSE.txt"
 servers:
   - url: "http://localhost:3000"
     description: "For local development/testing"
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -424,7 +424,7 @@ components:
       in: header
       name: X-Octobox-API
   schemas:
-    profile:
+    User:
       title: User Information
       type: object
       properties:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -593,16 +593,7 @@ components:
         subject: 
           "$ref": "#/components/schemas/subject"
         repo:
-          type: object
-          properties:
-            id:
-              type: integer
-            name:
-              type: string
-            owner:
-              type: string
-            repo_url:
-              type: string
+          "$ref": "#/components/schemas/repo"
 
     subject:
       title: Subject
@@ -617,6 +608,7 @@ components:
         state:
           type: string
     repo:
+      title: Repo
       type: object
       properties:
         id:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -562,16 +562,7 @@ components:
         subject:
           "$ref": "#/components/schemas/subject"
         repo:
-          type: object
-          properties:
-            id:
-              type: integer
-            name:
-              type: string
-            owner:
-              type: string
-            repo_url:
-              type: string
+          "$ref": "#/components/schemas/repo"
 
     subject-url-response:
       title: Subject URL Response
@@ -624,4 +615,15 @@ components:
         type:
           type: string
         state:
+          type: string
+    repo:
+      type: object
+      properties:
+        id:
+          type: integer
+        name:
+          type: string
+        owner:
+          type: string
+        repo_url:
           type: string
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1,0 +1,669 @@
+openapi: '3.0.2'
+info:
+  title: "Octobox"
+  description: "Octobox helps you manage your GitHub notifications efficiently so you can spend more time getting things done. It provides several features to manage GitHub notifications, including archived states, starred noticiations, extended filtering, and pinned searches."
+  version: "1" #unsure here
+  termsOfService: "https://octobox.io/terms"
+  contact:
+    name: "Octobox API"
+    url: "https:/octobox.io"
+    # email: ""
+  license:
+    name: "GNU AFFERO GENERAL PUBLIC LICENSE"
+    url: "https://github.com/octobox/octobox?tab=AGPL-3.0-1-ov-file#readme"
+servers:
+  - url: "http://localhost:3000"
+    description: "For local development/testing"
+  - url: "http://octobox.io" # Is this the correct server name?
+tags:
+  - name: Notifications
+    description: "Return and process notifications"
+  - name: Users
+    description: "Return and modify user profile information"
+  - name: Pinned Searches
+    description: "Create, return, and modify pinned searches"
+externalDocs:
+  description: Octobox Documentation
+  url: https://octobox.io/documentation
+security:
+  - BearerAuth: []
+  - ApiKeyAuth: []
+paths:
+  /api/notifications/archive_selected.json:
+    post:
+      tags:
+        - Notifications
+      summary: "Archive notifications."
+      description: "Archive GitHub notifications. You can choose to archive all notifications or only selected ones."
+      operationId: ArchiveNotifications
+      parameters:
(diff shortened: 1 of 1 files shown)


andrew:
Looking good, thanks for tackling this, I've just merged an update that fixes running octobox via docker-compose, so that may solve some of the issues you were running into.

rickwysocki:
Thanks for the update! I verified that Swagger's curl command works for `api/notifications/sync.json`, but I'm still getting responses of very large HTML/Ruby files when using the commands produced for syncing, posting pinned searches, and updating pinned searches. There don't seem to be any changes made to the local version of Octobox I'm running using Docker, either. I'll take another look tomorrow, but any advice would be helpful. 

I'm happy to work on this as long as is useful, by the way, as I'm trying to get my API documentation chops together.

andrew:
It could well be broken, although all the tests are passing which makes me think it’s a config issue, I’ll also give them a try tomorrow. 

This is really helpful, thanks for contributing

rickwysocki:
Hi there: Just an update that I tried those commands using my live Octobox URL rather than Docker. While this didn't return HTML in terminal, I still wasn't able to post or patch pinned notifications using CURL. Additionally, I'm also still not sure whether syncing feature is working or not.

rickwysocki:
Hi there! Just following up to ask if you'd had a chance to look at those endpoints and/or any instructions for to continuing working on the API documentation.

andrew:
Sorry for the delay, I’ll review this later tonight

openapi.yaml
```diff
@@ -0,0 +1,669 @@
...
+                properties:
+                  count:
+                    description: The mumber of unread notifications.
+                    type: number
+                    example: 1
+
+
+#PinnedSearchesController
+  /api/pinned_searches.json:
+    post: # I couldn't get this to work using CURL. Maybe it's because I'm using Docker/doing local testing? The JSON in the current documentation was throwing me a bit, too, and I had a hard time validating it. I don't think the example response is currently correct.
```
andrew, line 238: I was able to get this to work locally with curl: 
```
curl -X POST -d "pinned_search[query]=foo" -d "pinned_search[name]=bar" -H "Authorization: Bearer MYTOKEN" http://localhost:3000/api/pinned_searches.json
{"id":4,"user_id":1,"query":"inbox:true foo","name":"bar","count":6,"created_at":"2024-03-14T21:16:48.433Z","updated_at":"2024-03-14T21:16:48.433Z"}
```

openapi.yaml
```diff
@@ -0,0 +1,669 @@
...
+          schema:
+            type: number
+      responses:
+        200:
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/return-pinned-search'
+    patch: # I couldn't get this to work using CURL. Maybe it's because I'm using Docker/doing local testing? The JSON in the current documentation was throwing me a bit, too, and I had a hard time validating it. I don't think the example response is currently correct.
```
andrew, line 321: I managed to get it working locally with curl: 
```
curl -X PATCH -d "pinned_search[query]=foobar" -d "pinned_search[name]=baz" -H "Authorization: Bearer MYTOKEN" http://localhost:3000/api/pinned_searches/4.json
```

andrew (changes requested):

openapi.yaml
```diff
@@ -0,0 +1,669 @@
...
+  securitySchemes:
+    BearerAuth:
+      type: http
+      scheme: bearer
+    ApiKeyAuth:
+      type: apiKey
+      in: header
+      name: X-Octobox-API
+  schemas:
+    profile:
```
andrew, line 427:
```suggestion
    User:
```

openapi.yaml
```diff
@@ -0,0 +1,669 @@
...
+        created_at:
+          type: string
+          description: The date and time of creation.
+          example: "2017-02-22T15:49:32.104Z"
+        updated_at:
+          type: string
+          description: The date and time of the last update.
+          example: "2017-02-22T15:49:32.104Z"
+
+    patch-response:
```
andrew, line 456: Can we rename this to something like "Refresh setting"

openapi.yaml
```diff
@@ -0,0 +1,669 @@
...
+              type: integer
+        unread_repositories:
+          type: object
+          properties:
+            octobox/octobox:
+              type: integer
+        notifications:
+          type: array
+          items:
+            type: object
```
andrew, line 575: Can we pull this object out into it's own `Notification` Object and do:

```
items:
  "$ref": "#/components/schemas/Notification"
```

here instead

openapi.yaml
```diff
@@ -0,0 +1,669 @@
...
+        web_url:
+          type: string
+        last_read_at:
+          type: string
+        created_at:
+          type: string
+        updated_at:
+          type: string
+        subject:
+          type: object
```
andrew, line 649: Can we pull this out into it's own `Subject` object and reference it here

openapi.yaml
```diff
@@ -0,0 +1,669 @@
...
+          properties:
+            title:
+              type: string
+            url:
+              type: string
+            type:
+              type: string
+            state:
+              type: string
+        repo:
```
andrew, line 659: Can we pull this out into it's own `Repo` object and reference it here

andrew:
I've added some inline comments, it's looking good so far, the schemas could do with a bit of tweaking to seperate the different objects and use the `$ref` feature of openapi.

Example: https://github.com/ecosyste-ms/packages/blob/main/openapi/api/v1/openapi.yaml#L1103

I could get all of the api endpoints working, a couple return 200 instead of 204, the creating and updating of pinned searches is a bit confusing as it follows the rails convention for parameters, I left examples in the comments above.

andrew:
I also opened up the file in https://editor.swagger.io, all working and no errors!

Screenshot 2024-03-14 at 21-33-44 Swagger Editor

Gerard19742608jourdain:
simply space

rickwysocki:
Thanks for this feedback! It's really helpful on my end. I'll start working on your recommendations and check in once I've finished the revisions.

rickwysocki:
I've made some revisions to the `openapi.yaml` file based on your helpful recommendations. The only things I still seem to have issues with are the two sync-related endpoints. 

1. For the `api/notifications/sync.json` endpoint, I get a 200 response with no body in Swagger. In Terminal, I seem to be getting a 400 error when I run a verbose CURL request produced by Swagger.
2. For the `api/notifications/syncing.json` endpoint, I get a 200 response in both Swagger and Terminal, but it comes with an `"error": null`.

I've attached screenshots showing those errors on my end. Everything else seems to be working for me, and I've made your requested changes. 

------------

image

------------

image

rickwysocki:
Hi there: Just touching base to see whether there is more work I can do on this. No rush on my end!

andrew:
@rickwysocki nothing more needed from your right now, I'm travelling for work at the moment, hope to get back to this in about a week

andrew:
Sorry for the delay @rickwysocki, I've merged this now, thanks for your work on it.

PR #3873, opened 2024-09-25
Use luminance to determine label foreground color

pradyunsg:
Use the luminance values from WCAG 2.0 guidelines to determine which foreground colour to use for labels.

The referenced StackOverflow answer was updated to note that the older implementation has suboptimal contrast for some colours and uses a similar computation as well.


andrew (approved):

andrew:
Thanks!

PR #4030, opened 2025-02-04
Update Dockerfile to Multi-Stage

multi-stager:
Hi,

Thanks for maintaining octobox. I represent a research group investigating [multi-stage builds](https://docs.docker.com/build/building/multi-stage/). We recently refactored your Dockerfile to a multi-stage Dockerfile and found that it brings the following benefits: 

✅ Reduced final image size (from 994MB to 120MB)  
✅ Minimized image vulnerabilities (verified via [Trivy](https://github.com/aquasecurity/trivy), from 13 to 1)  

We hope this small improvement to the Dockerfile will be useful to you :)

Thanks,

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM ruby:3.3.6-alpine
+### Build stage
+FROM ruby:3.3.6-alpine AS builder
 
 ENV APP_ROOT /usr/src/app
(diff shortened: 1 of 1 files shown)


andrew:
this doesn't work as none of the ruby gems are available at runtime

PR #4511, opened 2026-05-13
fix: include updated_at in pluck for PostgreSQL DISTINCT compatibility

wedamija:
PostgreSQL enforces that ORDER BY expressions must appear in the SELECT list when using DISTINCT. The `newest` scope orders by `updated_at`, but `pluck(:id)` only selects `id`, causing PG::InvalidColumnReference errors on the notification show and expand_comments actions.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -18,7 +18,7 @@ class NotificationsController < ApplicationController
     scope = notifications_for_presentation.newest
     scope = load_and_count_notifications(scope) unless request.xhr?
 
-    ids = scope.pluck(:id)
+    ids = scope.pluck(:updated_at, :id).map(&:last)
     position = ids.index(params[:id].to_i)
     @notification = current_user.notifications.find(params[:id])
     @previous = ids[position-1] unless position.nil? || position-1 < 0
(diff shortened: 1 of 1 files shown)


andrew (changes requested):
Thanks for tracking this down. The change does fix the common case but I'd like to push back on the approach before merging.

The `.distinct` that triggers this comes from the `label` scope in `lib/octobox/notifications/inclusive_scope.rb`. By adding `updated_at` to the `pluck` here, we're teaching two call sites about an implementation detail of a scope they don't otherwise know about. The next place that plucks or selects from a label-filtered relation will hit the same error and need the same workaround.

It's also not a complete fix. `Search#apply_sort` can reorder by `upper(notifications.subject_title)` or `last_read_at` before `.newest` appends `updated_at`. A query like `label:bug sort:subject` still produces `SELECT DISTINCT updated_at, id ... ORDER BY upper(subject_title), updated_at` and PG rejects that for the same reason.

Could you instead change the `label` scope so it doesn't apply `.distinct` to the outer relation? Something like wrapping the labels join in a subquery on `notifications.id` (`where(id: ...)`) would dedupe without constraining what callers can select or order by, and these two controller actions could stay as `pluck(:id)`.

Please also add a test. A controller test for `show` with a `label:` filter (and ideally one combining `label:` with `sort:subject`) would have caught the original error and will stop it regressing.

Small note on the description: PG has always required ORDER BY columns to appear in the select list under `SELECT DISTINCT`, it isn't new in 16. Worth correcting so future readers don't go digging through the wrong release notes.

andrew:
Closing in favour of #4576, which fixes this in the `label` scope itself so it also covers the `label:` + `sort:subject` case and adds tests for both. Thanks again for the report — happy to reopen if you'd rather pick it back up.

PR #4569, opened 2026-06-30
Handle 423 Locked sync polling

knu:
Treat the `423 Locked` response from `/notifications/syncing.json` as a pending sync state instead of a completed sync.

The previous client code always parsed the response body and then reloaded the page when no error was present.  While a background sync was still working, the server returned `423 Locked` with an empty JSON body, which made the UI reload immediately and enter a visible refresh loop.

Poll again after a short delay while the sync is still locked, and only reload the page once the endpoint returns an ordinary completion response.

### DIFF
--- a/app/assets/javascripts/octobox.js
+++ b/app/assets/javascripts/octobox.js
@@ -331,8 +331,17 @@ var Octobox = (function() {
     }
 
     fetch("/notifications/syncing.json")
-    .then(response => response.json())
+    .then(response => {
+      if (response.status === 423) {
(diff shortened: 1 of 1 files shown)


andrew (approved):
Fixes a regression from 81344b1e where the jQuery → fetch conversion lost the 423 handling — `$.ajax` treated 423 as an error and rescheduled the poll, `fetch()` doesn't reject on HTTP status so it fell through to `Turbolinks.visit` and reload-looped while the sync job was running. Thanks!

PR #4570, opened 2026-06-30
Support large GitHub account IDs

knu:
There are existing accounts with bigint IDs, and Octobox cannot handle notifications containing them.


andrew (approved):
GitHub account IDs passed int32 max so `account_id`/`target_id` would overflow on newer accounts. Follows the same pattern as 20240221191013. Tables are small in production (~4k/~300/6 rows) so the `change_column` rewrites are sub-second. Thanks!

PR #4572, opened 2026-06-30
Fix false Sync failed alert

knu:
The sync endpoint starts a notification sync but does not return a JSON payload.  Treat an empty successful response as success, then poll syncing.json for the actual sync result.


andrew (approved):
Same root cause as #4569 — another jQuery→fetch regression from 81344b1e. `/notifications/sync.json` returns 204 with an empty body, so `response.json()` throws and the catch shows "Sync failed" even though the job was queued. Dropping the parse and checking `response.ok` is the right fix; `refreshOnSync` never used the argument anyway. Thanks for chasing all of these down!

